PR #1, opened 2015-10-12
DM-4001

jonathansick:
This PR for DM-4001 adds a section on writing stack documentation in the new Sphinx-based documentation architecture. I've included pages on the reStructuredText markup language, how to write package documentation, and how to write documentation embedded in Python and C++ source files.

See the documentation online at http://docs.lsst.codes/en/tickets-dm-4001/development/docs/index.html


jmatt:
A couple other comments.

If we expect the documentation to continue to change I think a reference to the documentation being a living document makes sense. Also is there is a release schedule for continued updated docs? What about references to related RFCs when changes are made. Maybe at the bottom as a link like what wikipedia does. I know any additional related links can add significant value. And related to that do we have a policy on linking to JIRA, RFCs, etc in code or in in docs to provide this sort of context?

jmatt:
+1

PR #12, opened 2016-01-26
[DM-4888] Update documentation to support git-lfs 1.1.0+.

jmatt:
* Reasoning: git-lfs 1.1.0+ doesn't accept empty username and passwords for the git-lfs server or secondary storage system. So the documentation needs to be updated.
* Add documentation to configure the git credential store to use empty username and passwords for the secondary storage systems and the git-lfs server.

	modified:   tools/git_lfs.rst

### DIFF
--- a/tools/git_lfs.rst
+++ b/tools/git_lfs.rst
@@ -24,12 +24,7 @@ Most package managers also provide the ``git-lfs`` client.
 Since, LFS is a rapidly evolving technology, package managers will help you keep up with new ``git-lfs`` releases.
 For example, Mac users with Homebrew_ can simply run ``brew install git-lfs`` and ``brew upgrade git-lfs``.
 
-.. _git-lfs-config:
-
-Configuring Git and Git LFS
-===========================
-
-Once ``git-lfs`` is installed, run
+Once ``git-lfs`` is installed, run:
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


jonathansick:
I've pushed a new commit based on this branch to [u/jonathansick/DM-4888](https://github.com/lsst-sqre/dm_dev_guide/commit/c4b33b3a6e383956ebbcb1bc181070aed88b2e58). As the commit message says, I think it clarifies what's going on for the reader. But I'm not 100% sure that it's the correct way of thinking about it, so I want to have you look at it before putting that commit on this branch.

See http://developer.lsst.io/en/u-jonathansick-dm-4888/tools/git_lfs.html

jmatt:
Jonathan, I like your approach. I'm going to cherry-pick it into this PR. Thanks again for the help.

tools/git_lfs.rst
```diff
@@ -31,20 +36,70 @@ Once ``git-lfs`` is installed, run
...
+
+.. code-block:: text
+   
+   https://:@lsst-sqre-prod-git-lfs.s3-us-west-2.amazonaws.com
+   https://:@s3.lsst.codes
+
+*Notice that the username is purposefully missing from these lines; this tells the Git credential store that you're opting for anonymous access.*
+
+Next, decide whether you will need to push Git LFS data, or only clone and pull form Git LFS managed repositories.
+This affects how you set up authentication to DM's Git LFS server (we only configured anonymous access to the *storage services* so far).
```
timj: "we have only configured" or "we only configure"

"storage services" suggests to me that you are only configuring anonymous access for pushes and not pulls. By "storage services" do you mean "services for retrieving data" rather than "services for storing data"?

jonathansick: The 'we' is well take.

@jmatt what's the appropriate name for the two types of servers (i.e., the S3 ones vs the `git-lfs.lsst.codes` ones)? At this point I simply call the S3 servers the 'storage services' and the `git-lfs.lsst.codes` server the Git LFS server.

jonathansick:
After talking to Josh and Tim, I think we've come up with a more thinking-free / copy-and-pastable treatment for this. Hold on while I implement and push it.

jonathansick:
Okay @jmatt checkout this [commit](https://github.com/lsst-sqre/dm_dev_guide/commit/9125c903b28c928c7e57691080e5a50c9262f78c) and see the rendering at http://developer.lsst.io/en/u-jonathansick-dm-4888/tools/git_lfs.html (Read the Docs is stalling on the build, so you won't see it there just yet).

Does this refactor work for you @timj?

jonathansick:
Next time it'd be nice to squash down a lot of the frivolous commits.

jonathansick:
Also I didn't finish the review.

timj:
I'm still not seeing the new version on that link.

jonathansick:
Yes, read the docs hasn't been able to successfully build any of the new versions this morning.

jmatt:
I usually don't squash, but I can clean it up if you want. I can't think of a time that squashing has been beneficial to me in my career. My review workflow doesn't require responsible committing. Too many junior devs and contracts. But it may make sense for a docs repository.

@jonathansick I apologize for rushing through this. I need documents I can reference so I can move git-lfs. We can keep improving the docs out-of-band.

PR #31, opened 2016-04-29
Update docs for third party packages

timj:


build-ci/new_package.rst
```diff
@@ -2,24 +2,35 @@
...
-`LSST organization on GitHub`_ and access must be granted to the appropriate
-teams. Both DM-written packages and third-party packages should go through the
-RFC process first to confirm their name and the suitability of their contents.
-
-DM-written Science Pipelines packages should follow the template in the
-`lsst/templates`_ repository.
-
-Third-party packages should be packaged as described in :doc:`third_party`.
+New packages intended for distribution to end users should generally be added
+as a dependency of a "top level products": these are the roots of the LSST
```
timj: "product" not "products".

build-ci/new_package.rst
```diff
@@ -2,24 +2,35 @@
...
+package hierarchy. They include ``lsst_apps``, ``lsst_distrib``,
```
timj: although `lsst_distrib` contains `lsst_apps` so it's hard to say `lsst_apps` is a root... We treat it as a root though.

build-ci/new_package.rst
```diff
@@ -2,24 +2,35 @@
...
 The new package must be added to the `etc/repos.yaml file in the lsstsw
-package`_ along with its corresponding GitHub URL. Note that this file is
-governed by a "self-merge" policy; see `RFC-75`_ for details.  The new package
-then needs to be added to the :file:`ups/*.table` file (and possibly the
-:file:`ups/*.cfg` file) of one or more other packages in the stack where it is
-used.
+package`_ along with its corresponding GitHub URL. This file is
+governed by a "self-merge" policy: upon opening a pull request, it will be
+checked by the :ref:`build-ci-travis` system, and developers may merge without
+further review on success. Refer to `RFC-75`_ for details.
```
timj: I don't think @ktlim approves of referencing RFCs for additional detail. I think he prefers that all the required detail is present in the approved document (the developer guide in this case) and that the RFC can be referred to for background information and history. Not detail. I will concede that you didn't actually write the phrase here...

build-ci/third_party.rst
```diff
@@ -251,7 +251,12 @@ an annotated tag to your package repository using::
 
     git tag -a versionNumber -m "Some comment."
 
-The initial versionNumber should match the external package's version number.
+The initial ``versionNumber`` should match the external package's version
+number, where available; otherwise, it is acceptable to generate a version
+string based on e.g. an upstream git SHA1 or revision number. Note that
+version numbers should not start with a letter, as EUPS regards that as
+semantically significant.
```
timj: Do we want to suggest a form? `0.SHA1.lsstN` ? Otherwise people might look at the SHA1, see that theirs does start with a number and then just use it directly. I do wonder if we shouldn't have something like `0.1.SHA1`, then `0.2.SHA1` for the next release (or drop the leading 0.), so that we can look at the version numbers and see an actual progression (and we can sort the tags numerically!)

build-ci/new_package.rst
jdswinbank: I thought I was doing a Good Thing here, by adding an explanation rather than just a reference to the RFC!

I'm in complete agreement with your (...K-T's) general point, but I don't want to start re-writing this text to include the screenshots etc available on the RFC within the scope of the current ticket.

timj: I was expecting you to say that. How about "Refer to RFC-75 for background."?

jdswinbank: Deal.

build-ci/third_party.rst
jdswinbank: Works for me; I've updated the text appropriately.

PR #33, opened 2016-05-05
Add Astropy usage recommendation

timj:
This is going to be a fluid section of the dev guide as we develop Astropy interoperability and integration.

### DIFF
--- a/coding/using_astropy.rst
+++ b/coding/using_astropy.rst
@@ -1,0 +1,43 @@
+#############
+Using Astropy
+#############
+
+.. _cpp_using_astropy:
+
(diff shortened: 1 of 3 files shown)


jonathansick:
See http://developer.lsst.io/en/tickets-dm-5934/coding/using_astropy.html

jonathansick:
@timj see the branch `tickets/DM-5934-intersphinx` for intersphinx usage. You'll want to cherry-pick the first commit on that branch; the second one is more of an example.

Going along with the commit message, this is what the links looks like:

napkin 16-05-10 10 56 06 am

For more on intersphinx see http://www.sphinx-doc.org/en/stable/ext/intersphinx.html

PR #36, opened 2016-05-26
DM-5973: Add python testing instructions

timj:
Modifications to the developer guide to add instructions on how LSST DM code should do python testing.

### DIFF
--- a/coding/python_testing.rst
+++ b/coding/python_testing.rst
@@ -4,25 +4,204 @@ Python Unit Testing
 
 This page provides technical guidance to developers writing unit tests for DM's Python code base.
 See :doc:`unit_test_policy` for an overview of LSST Stack testing.
+If you have legacy DM :mod:`unittest` ``suite``-based code (code that sets up a :class:`unittest.TestSuite` object by listing specific test classes and that uses :lfunc:`lsst.utils.tests.run` rather than :func:`unittest.main`) please refer to tech note `SQR-012`_ for porting instructions.
+LSST tests should be written using the :mod:`unittest` framework, with default test discovery, and should support being run using the `pytest`_ test runner.
 
-Memory Leak Testing
-===================
+.. _SQR-012: http://sqr-012.lsst.io
+.. _pytest: http://pytest.org
 
-:lmod:`lsst.utils.tests` provides several utilities for writing Python tests
-that developers should make use of. In particular,
-:lclass:`lsst.utils.tests.MemoryTestCase` is used to detect memory leaks in
-C++ objects. :lclass:`~lsst.utils.tests.MemoryTestCase` should be used in
-*all* tests, even if C++ code is not explicitly referenced.
+Introduction to ``unittest``
+============================
+
+This document will not attempt to explain full details of how to use :mod:`unittest` but instead shows common scenarios encountered in the LSST codebase.
+
+A simple :mod:`unittest` example is shown below:
+
+.. literalinclude:: unit_test_snippets/unittest_basic_example.py
+   :language: python
+
+The important things to note in this example are:
+
+* If the test is being executed using :command:`python` from the command line the :py:func:`unittest.main` call performs the test discovery and executes the tests, setting exit status to non-zero if any of the tests fail.
+* Test classes are executed in the order in which they appear in the test file.
+  In this case the tests in ``DemoTestCase1`` will be executed before those in ``DemoTestCase2``.
+* Test classes must, ultimately, inherit from :class:`unittest.TestCase` in order to be discovered and it is :ref:`recommended <py-utils-tests>` that :lclass:`lsst.utils.tests.TestCase` be used as the base class when :lmod:`~lsst.afw` objects are involved.
+  The tests themselves must be methods of the test class with names that begin with ``test``.
+  All other methods and classes will be ignored by the test system but can be used by tests.
+* If a test method completes the test passes, if it throws an uncaught exception the test has failed.
+
+Using a more advanced test runner
+=================================
+
+All tests should be written such that they are runnable using `pytest`_, a policy adopted in :jira:`RFC-69`.
+`pytest`_ provides a much richer execution and reporting environment for tests and can be used to run multiple tests files together.
+`pytest`_ test discovery is much more flexible than that provided by :mod:`unittest` but LSST test files should not take advantage of that flexibility as it can lead to inconsistency in test reports that depend on the specific test runner.
+In particular, care must be taken not to have free functions that use a ``test`` prefix or non-\ :class:`~unittest.TestCase` test classes that are named with a ``Test`` prefix in the test files.
+
+.. note::
+   The normal way to run tests is to use :command:`scons` or :command:`scons tests`.
+   :command:`scons` will then execute each of the tests and report any failures.
+   Currently tests are executed one at a time using :command:`python` and checking the command exit status.
+   Eventually :lmod:`~lsst.sconsUtils` will be modified to call :command:`py.test` directly with all Python test files (possibly with a different order each invocation).
+   It is therefore important that during this transition period developers check that tests run correctly with `pytest`_ by explicitly running :command:`py.test` when creating or updating test files.
+
+`pytest`_ can run tests from more than one file in a single invocation and this can be used to verify that there is no state contaminating later tests.
+To run `pytest`_ use the :command:`py.test` executable:
+
+.. code-block:: shell
+
+   $ py.test tests/*.py
+
+and to ensure that the order of test execution does not matter it is useful to sometimes run the tests in reverse order:
+
+.. code-block:: shell
+
+   $ py.test `ls -r tests/*.py`
+
+When writing tests it is important that tests are skipped using the proper :mod:`unittest` :ref:`skipping framework <python:unittest-skipping>` rather than returning from the test early.
+:mod:`unittest` supports skipping of individual tests and entire classes using decorators or skip exceptions.
+LSST code sometimes raises skip exceptions in :meth:`~unittest.TestCase.setUp` or :meth:`~unittest.TestCase.setUpClass` class methods and :mod:`unittest` provides much flexibility in how skipping can be implemented.
+It is also possible to indicate that a particular test is expected to fail, being reported as an error if the test unexpectedly passes.
+Expected failures can be used to write test code that triggers a reported bug before the fix to the bug has been implemented and without causing the continuous integration system to die.
+One of the primary advantages of using a modern test runner such as `pytest`_ is that it is very easy to generate machine-readable pass/fail/skip/xfail statistics to see how the system is evolving over time.
+
+Using a shared base class
+=========================
+
+For some tests it is helpful to provide a base class and then share it amongst multiple test classes that are configured with different attributes.
+If this is required be careful to not have helper functions prefixed with ``test`` and do not have the base class named with a ``Test`` prefix and ensure it does not inherit from :class:`~unittest.TestCase`, if you do `pytest`_ will attempt to find tests inside it.
+Historically LSST code has dealt with this by creating a test suite that only includes the classes to be tested, omitting the base class.
+This does not work in a `pytest`_ environment.
+
+Consider the following test code:
+
+.. literalinclude:: unit_test_snippets/unittest_baseclass.py
+   :language: python
+
+which inherits from the helper class and :class:`unittest.TestCase` and runs a single test without attempting to run any tests in ``BaseClass``.
+
+.. code-block:: shell
+
+   $ py.test -v coding/unit_test_snippets/unittest_baseclass.py
+   ======================================= test session starts ========================================
+   platform darwin -- Python 3.4.3, pytest-2.9.1, py-1.4.30, pluggy-0.3.1 -- /usr/local/bin/python3.4
+   cachedir: coding/unit_test_snippets/.cache
+   rootdir: coding/unit_test_snippets, inifile:
+   collected 1 items
+
+   coding/unit_test_snippets/unittest_baseclass.py::ThisIsTest1::testParam PASSED
+
+   ===================================== 1 passed in 0.02 seconds =====================================
+
+
+LSST Utility Test Support Classes
+=================================
+
+:lmod:`lsst.utils.tests` provides several utilities functions and classes for writing Python tests that developers should make use of.
+
+.. _py-utils-tests:
+
+Special Asserts
+---------------
+
+Inheriting from :lclass:`lsst.utils.tests.TestCase` rather than :class:`unittest.TestCase` enables two new asserts that are useful for doing element-wise comparison of two :mod:`numpy` arrays.
+
+:lmeth:`~lsst.utils.tests.TestCase.assertClose`
+   Asserts that the corresponding elements in two arrays are within the specified tolerance.
+:lmeth:`~lsst.utils.tests.TestCase.assertNotClose`
+   Asserts that the two arrays are different.
+
+Additionally, :lmod:`~lsst.afw` provides additional asserts that get loaded into :lclass:`lsst.utils.tests.TestCase` when the associated module is loaded.
+
+:lmeth:`~lsst.afw.coord.utils.assertCoordsNearlyEqual`
+   Assert that two coords represent nearly the same point on the sky (provided by :lmod:`lsst.afw.coord.utils`).
+:lmeth:`~lsst.afw.geom.utils.assertAnglesNearlyEqual`
+   Assert that two angles are nearly equal, ignoring wrap differences by default (provided by :lmod:`lsst.afw.geom.utils`).
+:lmeth:`~lsst.afw.geom.utils.assertPairsNearlyEqual`
+   Assert that two planar pairs (e.g. :lclass:`~lsst.afw.geom.Point2D` or :lclass:`~lsst.afw.geom.Extent2D`) are nearly equal (provided by :lmod:`lsst.afw.geom.utils`).
+:lmeth:`~lsst.afw.geom.utils.assertBoxesNearlyEqual`
+   Assert that two boxes (:lclass:`~lsst.afw.geom.Box2D` or :lclass:`~lsst.afw.geom.Box2I`) are nearly equal (provided by :lmod:`lsst.afw.geom.utils`).
+:lmeth:`~lsst.afw.image.basicUtils.assertWcsNearlyEqualOverBBox`
+   Compare :lmeth:`~lsst.afw.image.imageLib.Wcs.pixelToSky` and :lmeth:`~lsst.afw.image.imageLib.Wcs.skyToPixel` for two WCS over a rectangular grid of pixel positions (provided by :lmod:`lsst.afw.image.basicUtils`).
+:lmeth:`~lsst.afw.image.testUtils.assertImagesNearlyEqual`
+   Assert that two images are nearly equal, including non-finite values (provided by :lmod:`lsst.afw.image.testUtils`).
+:lmeth:`~lsst.afw.image.testUtils.assertMasksEqual`
+   Assert that two masks are equal (provided by :lmod:`lsst.afw.image.testUtils`).
+:lmeth:`~lsst.afw.image.testUtils.assertMaskedImagesNearlyEqual`
+   Assert that two masked images are nearly equal, including non-finite values (provided by :lmod:`lsst.afw.image.testUtils`).
+
+Testing Executables
+-------------------
+
+In some cases the test to be executed is a shell script or a compiled binary executable.
+In order for the test running environment to be aware of these tests a Python test file must be present that can be run by `pytest`_ to executes these tests in the :mod:`unittest` environment.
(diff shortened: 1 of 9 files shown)


coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
 
 This page provides technical guidance to developers writing unit tests for DM's Python code base.
 See :doc:`unit_test_policy` for an overview of LSST Stack testing.
+If you have legacy DM :mod:`unittest` ``suite``-based code please refer to tech note `SQR-012`_ for porting instructions.
```
r-owen: A few more words about what this means could help ("suite-based" isn't very clear). I am pretty sure you mean our tests that create a test suite and then call it, but it may be best to say so or include a representative snippet (something that makes it easy to tell if a test file is "old style").

timj: Yes, I was trying not to spend time documenting a scheme that we had inherited through cargo-culting rather than documented policy. I'll have a go at putting enough words in to help people spot the old style without documenting the old style here (SQR-012 does describe the old and the new).

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+
+.. literalinclude:: unit_test_snippets/unittest_basic_example.py
+   :language: python
+
+The important things to note in this example are:
+
+* If the test is being executed using ``python`` from the command line the :py:func:`unittest.main` call performs the test discovery and executes the tests, setting exit status to non-zero if any of the tests fail.
+* Test classes are executed in the order in which they appear in the test file.
+  In this case the tests in :class:`DemoTestCase1` will be executed before those in :class:`DemoTestCase2`.
+* Test classes must, ultimately, inherit from :class:`unittest.TestCase` in order to be discovered, and tests themselves must be methods that start with ``test``.
```
r-owen: Ultimately, yes, but LSST recommends inheriting from lsst.utils.tests.TestCase so could you please list that first? That way we train folks right away.

timj: I'm not sure I agree that we recommend it. You need to use it if you want the `numpy` comparison asserts but why would we need to insist on always using it as the base test class?

r-owen:
I would say that utils.tests.TestClass is the recommended class for any package at the level of afw or higher, because so many tests benefit from the many useful asserts we add. Tests grow over time and it's a hassle to have to switch the base class because a standard (for LSST) assert is missing. If you disagree then we ought to consider widening this conversation -- at least to include K-T. Having seen too many comparisons that would have been done better by our asserts, I feel quite strongly we under-utilize that class!

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+* If a test method completes the test passes, if it throws an uncaught exception the test has failed.
+
+Using a more advanced test runner
+=================================
+
+All tests should be written such that they are runnable using `pytest`_, a policy adopted in `RFC-69`_.
+`pytest`_ provides a much richer execution and reporting environment for tests and can be used to run multiple tests files together.
+`pytest`_ test discovery is much more flexible than that provided by :mod:`unittest` but LSST test files should not take advantage of that flexibility as it can lead to inconsistency in test reports that depend on the specific test runner.
+In particular, care must be taken not to have free functions that use a ``test`` prefix or non-\ :class:`~unittest.TestCase` test classes that are named with a ``Test`` prefix in the test files.
+
```
r-owen: Ouch. Any hope of detecting and rejecting such things automatically?

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+.. _RFC-69: https://jira.lsstcorp.org/browse/RFC-69
+
+`pytest`_ can run tests from more than one file in a single invocation and this can be used to verify that there is no state contaminating later tests.
+To run `pytest`_ use the :command:`py.test` executable:
```
r-owen: I suggest listing the standard way to run tests first ("scons" or "scons tests") and then point out this is how to run a subset of tests.

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+   $ py.test tests/*.py
+
+and to ensure that the order of test execution does not matter it is useful to sometimes run the tests in reverse order:
+
+.. code-block:: shell
+
+   $ py.test `ls -r tests/*.py`
+
+When writing tests it is important that tests are skipped using the proper :mod:`unittest` :ref:`skipping framework <python:unittest-skipping>` rather than returning from the test early.
+:mod:`unittest` can support skipping of individual tests and entire classes using decorators, and exceptions can also be used.
```
r-owen: can support -> supports. What does ", and exceptions can also be used" mean? I suggest this be a separate sentence with a bit more explanation.

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+LSST code sometimes raises skip exceptions in :meth:`~unittest.TestCase.setUp` or :meth:`~unittest.TestCase.setUpClass` class methods and :mod:`unittest` provides much flexibility in how skipping can be implemented.
+It is also possible to indicate that a particular test is expected to fail, to the extent that it is reported it a test unexpectedly passes.
```
r-owen: "that it is reported it a test..."? Please reword. Perhaps " "and it is reported as an error if the test unexpected passes"

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+Expected failures can be used to write test code that triggers a reported bug before the fix to the bug has been implemented and without causing the continuous integration system to die.
+One of the advantages of using a modern test runner such as `pytest`_ is that it is very easy to generate machine-readable pass/fail/skip/xfail statistics to see how the system is evolving over time.
```
r-owen: Isn't this our primary reason for adopting it? I suppose we also benefit from reduced boilerplate, but that never seemed burdensome to me.

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+
+Using a shared base class
+=========================
+
+For some tests it is helpful to provide a base class and then share it amongst multiple test classes that are configured with different attributes.
+If this is required be careful to not have helper functions prefixed with ``test`` and do not have the base class named with a ``Test`` prefix and ensure it does not inherit from :class:`~unittest.TestCase`, if you do `pytest`_ will attempt to find tests inside it.
+Historically LSST code has dealt with this by creating a test suite that only includes the classes to be tested, omitting the base class.
+This can not work in a `pytest`_ environment, so instead multiple inheritance should be used such that the test classes inherit from :class:`~unittest.TestCase` and the support base class.
```
r-owen: I think it would suffice to say "omitting the base class. This does not work in a pytest environment". You already explained what to do instead.

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+=================================
+
+:mod:`lsst.utils.tests` provides several utilities functions and classes for writing Python tests that developers should make use of.
+
+Special Asserts
+---------------
+
+Inheriting from :class:`lsst.utils.tests.TestCase` rather than :class:`unittest.TestCase` enables two new asserts that are useful for doing element-wise comparison of two :mod:`numpy` arrays.
+:meth:`~lsst.utils.tests.TestCase.assertClose` asserts that the corresponding elements in two arrays are within the specified tolerance.
+:meth:`~lsst.utils.tests.TestCase.assertNotClose` asserts that the two arrays are different.
```
r-owen: assertBoxEqual, assertMaskedImagesNearlyEqual, assertAnglesNearlyEqual, assertWcsNearlyEqual... there are a host of these, many of which are a pain to implement properly (to the extent that they often were NOT implemented properly).

coding/python_testing.rst
r-owen: It may be worth adding a forward reference to running non-python unit tests here (e.g. "discussed below" or "for information about..."

coding/python_testing.rst
```diff
@@ -4,6 +4,156 @@ Python Unit Testing
...
+
+.. literalinclude:: unit_test_snippets/unittest_runner_example.py
+   :linenos:
+   :language: python
+   :emphasize-lines: 6, 16, 21
+
+Note that :class:`~lsst.utils.tests.MemoryTestCase` must always be the
+final test suite.
+For the memory test to function properly the :func:`lsst.utils.tests.init()` function must be invoked before any of the tests in the class are executed.
+Since LSST test scripts are required to run properly from the command-line and when called from within `pytest`_, the :func:`~lsst.utils.tests.init()` function has to be called twice: once in the :func:`setup_module` function that is called by `pytest`_ whenever a test module is loaded, and also just before the call to :func:`unittest.main()` call.
```
r-owen: I find this is confusing. I think you are just warning us that it will automatically be called twice py pytest if we follow your rules. Surely it only has to appear once in each test file? If so, please clarify the text; perhaps turn it into a parenthetical remark.

coding/python_testing.rst
timj: I don't think pytest can be configured to solely use unittest default test discovery. It will be somewhat obvious when the person writes the tests so I don't think it's a problem once we have a stack that works with pytest. There is going to be work required to transition our test suites.

coding/python_testing.rst
timj: Yes it is. I don't think it hurts to reiterate best practice.

coding/unit_test_snippets/unittest_basic_example.py
```diff
@@ -0,0 +1,21 @@
+import unittest
+
+
+class DemoTestCase1(unittest.TestCase):
+    """Demo test case."""
+
+    def testDemo(self):
+        self.assertTrue(True)
```
r-owen: assertTrue is used far too heavily in our code. I beg you to replace assertTrue and assertFalse in all your examples with a something more generally useful. I suggest assertEqual as the first choice, then throw in something like assertLess (since it suggests the availability of many similar asserts). The other two I suggest you demo are assertAlmostEqual and `with self.assertRaises(...):` Those give a really good overview of the good stuff.

coding/python_testing.rst
r-owen: I agree. I think the sentence would be stronger if you added "primary": "One of the primary advantages of using...".

coding/python_testing.rst
r-owen: I don't know what you mean by "it will be somewhat obvious when the person writes the tests", but I'm not very worried about it. I expect a few will slip through our review process. If scons uses pytest to run our tests then it will not matter in practice.

coding/python_testing.rst
timj: I think I'm trying to explain why the testing code needs to call `init()` in two different places. One is to allow pytest to run the initialization code and the other is to allow the initialization code to run when the python script is executed from the command line by `python`.

coding/python_testing.rst
timj: The plan is to use pytest to run the tests but there will be a transitional period. I imagine that tests will continue to break under pytest until we make that switch but we can't do that change until we know we are mostly there (so I'll probably have a sconsUtils branch that uses pytest on Jenkins and try it to see how far things get). Note that the sconsUtils pytest change got bumped from F16 plan though and I don't think any work has been scheduled on fixing up all the breakage in meas_astrom and afw when pytest is used.

coding/python_testing.rst
timj: I had missed those because they are scattered across other packages and I had not expected that sort of monkey patching of the utils test case class. I am pondering how best to sort them out for documentation (they magically appear in utils when you import some chunk of AFW)

r-owen: Yes, just like functors that register themselves in a registry. Importing afw will make these particular asserts appear. Since they test afw classes they live in afw. It's not an ideal system, but it makes sense. If you are trying to test afw objects changes are you have imported afw (directly or indirectly). I'm not sure if any others are added; certainly afw is likely to do the most because it has lots of smallish objects that can usefully be compared, but are not trivial to compare.

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
 
 This page provides technical guidance to developers writing unit tests for DM's Python code base.
 See :doc:`unit_test_policy` for an overview of LSST Stack testing.
+If you have legacy DM :mod:`unittest` ``suite``-based code (code that sets up a :class:`unittest.TestSuite` object by listing specific test classes and that uses :lfunc:`lsst.utils.tests.run` rather than :func:`unittest.main`) please refer to tech note `SQR-012`_ for porting instructions.
```
ktlim: I'd usually put a comma before "please".

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+
+.. literalinclude:: unit_test_snippets/unittest_basic_example.py
+   :language: python
+
+The important things to note in this example are:
+
+* If the test is being executed using :command:`python` from the command line the :py:func:`unittest.main` call performs the test discovery and executes the tests, setting exit status to non-zero if any of the tests fail.
+* Test classes are executed in the order in which they appear in the test file.
+  In this case the tests in ``DemoTestCase1`` will be executed before those in ``DemoTestCase2``.
+* Test classes must, ultimately, inherit from :class:`unittest.TestCase` in order to be discovered and it is :ref:`recommended <py-utils-tests>` that :lclass:`lsst.utils.tests.TestCase` be used as the base class when :lmod:`~lsst.afw` objects are involved.
```
ktlim: Comma between "discovered" and "and".

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+  The tests themselves must be methods of the test class with names that begin with ``test``.
+  All other methods and classes will be ignored by the test system but can be used by tests.
+* If a test method completes the test passes, if it throws an uncaught exception the test has failed.
```
ktlim: Commas after "completes" and "exception"; existing comma should be a semicolon?

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+
+Using a more advanced test runner
+=================================
+
+All tests should be written such that they are runnable using `pytest`_, a policy adopted in :jira:`RFC-69`.
```
ktlim: Render has a strange "and" after "RFC-69".

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+`pytest`_ provides a much richer execution and reporting environment for tests and can be used to run multiple tests files together.
+`pytest`_ test discovery is much more flexible than that provided by :mod:`unittest` but LSST test files should not take advantage of that flexibility as it can lead to inconsistency in test reports that depend on the specific test runner.
```
ktlim: Comma before "but".

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+In particular, care must be taken not to have free functions that use a ``test`` prefix or non-\ :class:`~unittest.TestCase` test classes that are named with a ``Test`` prefix in the test files.
+
+.. note::
+   The normal way to run tests is to use :command:`scons` or :command:`scons tests`.
+   :command:`scons` will then execute each of the tests and report any failures.
+   Currently tests are executed one at a time using :command:`python` and checking the command exit status.
+   Eventually :lmod:`~lsst.sconsUtils` will be modified to call :command:`py.test` directly with all Python test files (possibly with a different order each invocation).
+   It is therefore important that during this transition period developers check that tests run correctly with `pytest`_ by explicitly running :command:`py.test` when creating or updating test files.
```
ktlim: Is `pytest` installed by default? Should there be a link to installation instructions?

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+
+and to ensure that the order of test execution does not matter it is useful to sometimes run the tests in reverse order:
+
+.. code-block:: shell
+
+   $ py.test `ls -r tests/*.py`
+
+When writing tests it is important that tests are skipped using the proper :mod:`unittest` :ref:`skipping framework <python:unittest-skipping>` rather than returning from the test early.
+:mod:`unittest` supports skipping of individual tests and entire classes using decorators or skip exceptions.
+LSST code sometimes raises skip exceptions in :meth:`~unittest.TestCase.setUp` or :meth:`~unittest.TestCase.setUpClass` class methods and :mod:`unittest` provides much flexibility in how skipping can be implemented.
```
ktlim: The part including and after "and unittest" seems somewhat redundant and could be dropped.

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+It is also possible to indicate that a particular test is expected to fail, being reported as an error if the test unexpectedly passes.
+Expected failures can be used to write test code that triggers a reported bug before the fix to the bug has been implemented and without causing the continuous integration system to die.
+One of the primary advantages of using a modern test runner such as `pytest`_ is that it is very easy to generate machine-readable pass/fail/skip/xfail statistics to see how the system is evolving over time.
+
+Using a shared base class
+=========================
+
+For some tests it is helpful to provide a base class and then share it amongst multiple test classes that are configured with different attributes.
+If this is required be careful to not have helper functions prefixed with ``test`` and do not have the base class named with a ``Test`` prefix and ensure it does not inherit from :class:`~unittest.TestCase`, if you do `pytest`_ will attempt to find tests inside it.
```
ktlim: Comma before "be careful".  Replace the first "and" with a period. Add a comma between "prefix" and "and".  Replace the comma with a semicolon.  Add a comma between "do" and "pytest".

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+Historically LSST code has dealt with this by creating a test suite that only includes the classes to be tested, omitting the base class.
+This does not work in a `pytest`_ environment.
+
+Consider the following test code:
+
+.. literalinclude:: unit_test_snippets/unittest_baseclass.py
```
ktlim: Shouldn't `BaseClass` derive from `object`?

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+
+   coding/unit_test_snippets/unittest_baseclass.py::ThisIsTest1::testParam PASSED
+
+   ===================================== 1 passed in 0.02 seconds =====================================
+
+
+LSST Utility Test Support Classes
+=================================
+
+:lmod:`lsst.utils.tests` provides several utilities functions and classes for writing Python tests that developers should make use of.
```
ktlim: "utility functions and classes" or "utilities, functions, and classes"?

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+:lmeth:`~lsst.afw.image.testUtils.assertMasksEqual`
+   Assert that two masks are equal (provided by :lmod:`lsst.afw.image.testUtils`).
+:lmeth:`~lsst.afw.image.testUtils.assertMaskedImagesNearlyEqual`
+   Assert that two masked images are nearly equal, including non-finite values (provided by :lmod:`lsst.afw.image.testUtils`).
+
+Testing Executables
+-------------------
+
+In some cases the test to be executed is a shell script or a compiled binary executable.
+In order for the test running environment to be aware of these tests a Python test file must be present that can be run by `pytest`_ to executes these tests in the :mod:`unittest` environment.
```
ktlim: Semi-optional comma between "tests" and "a" to break up the long sentence.  "executes" -> "execute".  "in the unittest environment" is somewhat redundant with "In order for the test running environment to be aware".

coding/python_testing.rst
```diff
@@ -4,6 +4,204 @@ Python Unit Testing
...
+In some cases an explicit test has to be written either because some precondition has to be met before the test will stand a chance of running or because some arguments have to be passed to the executable.
+To support this the :lmeth:`~lsst.utils.tests.ExecutableTestCase.assertExecutable` method is available:
+
+.. code-block:: python
+
+   def testBinary(self):
+       self.assertExecutable("binary1", args=None,
+                             root_dir=os.path.dirname(__file__))
+
+Where ``binary1`` is the name of the executable relative to the root directory specified in the ``root_dir`` optional argument.
```
ktlim: "Where" -> "where".

coding/python_testing.rst
timj: Good point. This is a conda package that needs to be installed. Obviously when we finally depend on pytest we wil add it to the dependencies properly.

coding/python_testing.rst
timj: Yes, although for explanation you can look at line 87 below...

coding/python_testing.rst
timj: @jonathansick am I using the `:jira:` syntax incorrectly?

jonathansick: 🤔 Investigating now.

jonathansick: @timj If you re-start the Travis build, and locally `pip install -U documenteer` you'll get a new version 0.1.7 that fixes this. I'm still trying to understand how the unit test did not catch exactly this...

timj: That worked. Thanks.

PR #39, opened 2016-06-01
Update Boost section to std default policy adopted in RFC-185

ktlim:


coding/using_boost.rst
```diff
@@ -4,24 +4,38 @@ Using Boost
 
 .. _cpp_using_boost:
 
-Certain Boost libraries are recommended; they should be used whenever applicable in preference to any other method of accomplishing the same effect. In general, any library that is tagged with *"Standard .... {something}"* in the `Boost library listing <http://www.boost.org/doc/libs>`_ falls into this category, unless the feature is already supported by the currently-mandated LSST ``g++`` version.
-Among others, this category includes:
+A Boost library shall be used in in preference to any other method of accomplishing the same effect only if:
```
ktlim: "in in", but I think everything between "used" and "only if" should be omitted.

coding/using_boost.rst
```diff
@@ -4,24 +4,38 @@ Using Boost
...
 
-* ``current_function``
-* ``format``
-* ``regex`` (until we adopt ``gcc`` 4.9 as our minimum compiler version, at which point use ``std::regex``)
-* ``test``
+1. the effect cannot be accomplished with a C++11 standard language feature, and
+2. a C++ standard library equivalent is not available, or
+3. the C++ standard library equivalent is not usable with our minimum supported compiler version (i.e. ``gcc`` version 4.8).
```
ktlim: Precedence of "and" and "or" here is a little unclear :).  Adding "standard language feature" to point 3 might remove some ambiguity.

coding/using_boost.rst
```diff
@@ -4,24 +4,38 @@ Using Boost
...
 
-Libraries that were accepted before we switched to C++11 (in particular made ``gcc`` 4.8 our minimum supported compiler), but have been or are being replaced by ``std::`` equivalents include:
+In particular the following libraries were accepted before we adopted ``gcc`` 4.8 as our minimum compiler version, but have been or are being replaced by ``std::`` equivalents,
+and are thus no longer accepted, include:
```
ktlim: "In particular, the following Boost libraries are no longer accepted as they have standard equivalents in gcc 4.8 and above:"

coding/using_boost.rst
```diff
@@ -4,24 +4,38 @@ Using Boost
...
 * ``type_traits``: use ``<type_traits>``
-* ``unordered_map``: use ``std::unordered_map`` from ``<unordered_map>``
+* ``unordered_map``: use ``<unordered_map>``
+
+Certain Boost libraries are recommended; they should be used whenever applicable in preference to any other method of accomplishing the same effect. In general, any library that is tagged with *"Standard .... {something}"* in the `Boost library listing <http://www.boost.org/doc/libs>`_ falls into this category (unless its use conflicts with the above rules on the availability of standard library / language equivalents).
+Among others, this category includes:
+
+* ``current_function``
+* ``format``
+* ``regex`` (until we adopt ``gcc`` 4.9 as our minimum compiler version, at which point use ``std::regex``)
```
ktlim: I thought @fritzm still had potential problems with `std::regex` due to lack of features.

coding/using_boost.rst
jdswinbank: Is the logic here correct? As I read what you've written, if the three criteria below are satisfied you *should* use Boost, even if other libraries which might do a better job (but which aren't in the standard) are available. I think you probably mean "may" rather than "shall".

coding/using_boost.rst
jdswinbank: I don't think I'd just tweak the words in point 3 here -- I don't think you can sensibly have a list like this which mixes its "and"s and it's "or"s without being confusing.

coding/using_boost.rst
```diff
@@ -4,24 +4,38 @@ Using Boost
...
+* ``ref``: use ``std::forward`` and universal / forwarding references
 * ``smart_ptr``: use ``std::shared_ptr`` and ``std::unique_ptr`` from ``<memory>`` instead of ``boost::shared_ptr`` and ``boost::scoped_ptr``
-* ``static_assert``: use ``static_assert``
+* ``static_assert``: use C++11 ``static_assert``
+* ``tuple``: use ``<tuple>``
 * ``type_traits``: use ``<type_traits>``
-* ``unordered_map``: use ``std::unordered_map`` from ``<unordered_map>``
+* ``unordered_map``: use ``<unordered_map>``
+
+Certain Boost libraries are recommended; they should be used whenever applicable in preference to any other method of accomplishing the same effect. In general, any library that is tagged with *"Standard .... {something}"* in the `Boost library listing <http://www.boost.org/doc/libs>`_ falls into this category (unless its use conflicts with the above rules on the availability of standard library / language equivalents).
```
jdswinbank: I'd say you want a colon, not a semi, here. And maybe this is the explanation of the "shall" above -- except that this seems to be saying you are only required to use a subset of Boost, not all of it.

coding/using_boost.rst
ktlim: Once upon a time we thought that something available in Boost was typically better than any competitor, if for no other reason than we were already depending on Boost.  Perhaps that is no longer true.

Also, my suggested rewrite doesn't have the same implication that Boost is preferable to anything else.

coding/using_boost.rst
jdswinbank: DM-3920 implies that Fritz would be happy with cutting over to `std::regex` in gcc 4.9. That said, I'm not sure I'd encode that as policy in this document -- who knows what else might change before we were ready to switch compiler?

coding/using_boost.rst
```diff
@@ -4,24 +4,38 @@ Using Boost
...
-* ``test``
+1. the effect cannot be accomplished with a C++11 standard language feature, and
+2. a C++ standard library equivalent is not available, or
+3. the C++ standard library equivalent is not usable with our minimum supported compiler version (i.e. ``gcc`` version 4.8).
 
-Libraries that were accepted before we switched to C++11 (in particular made ``gcc`` 4.8 our minimum supported compiler), but have been or are being replaced by ``std::`` equivalents include:
+In particular the following libraries were accepted before we adopted ``gcc`` 4.8 as our minimum compiler version, but have been or are being replaced by ``std::`` equivalents,
+and are thus no longer accepted, include:
 
 * ``array``: use ``std::array`` from ``<array>``
```
jdswinbank: This seems like it would be nicer as a definition list, but I guess you're just following the existing style -- probably not worth the churn to change it.

coding/using_boost.rst
ktlim: 👍 for leaving out the parenthetical.  In [this comment](https://jira.lsstcorp.org/browse/RFC-185?focusedCommentId=48240&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-48240), Fritz says that there are some things missing in `std::regex`, although as you say DM-3920 indicates that it is sufficient for `qserv` needs.

fritzm: Some time back I really tried to give std::regex a go.  Beyond the gcc 
4.8 bustage, I also found it to be missing some very useful things that 
I had liked with boost::regex:

  * named sub-expressions
  * look-behind
  * comments
  * PCRE compatibility in general (std::regex is basically ECMA)

Disappointing :-(

coding/using_boost.rst
pschella: Adjusted according to suggestions.

coding/using_boost.rst
pschella: I merged items two and three instead. (please see soon to be updated PR).

coding/using_boost.rst
pschella: Changed accordingly.

coding/using_boost.rst
pschella: I removed the sentence between braces. We can think about it later when we switch compilers.

coding/using_boost.rst
pschella: Agreed, and changed.

coding/using_boost.rst
pschella: Changed to a definition list. Let me know if it looks better or worse.

coding/using_boost.rst
```diff
@@ -4,26 +4,67 @@ Using Boost
...
+        use ``<cstdint>``
+
+``filesystem``
+        use ``<filesystem>``
+
+``lambda``
+        use C++11 lambda functions
+
+``lexical_cast``
+        use ``std::to_string``, ``std::sto``, ``std::stod`` etc.
```
jdswinbank: Should ``std::sto`` be ``std::stoi``?

coding/using_boost.rst
```diff
@@ -4,26 +4,67 @@ Using Boost
...
+1. the desired effect cannot be accomplished with a C++11 standard language feature, and
+2. a C++ standard library equivalent is either unavailable or unusable with our minimum supported compiler version (i.e. ``gcc`` version 4.8).
+
+In particular, the following Boost libraries are no longer accepted as they have standard equivalents in gcc 4.8 and above:
+
+``array``
+        use ``<array>``
+
+``bind``
+        preffer C++11 lambda functions instead, but use ``std::bind`` from ``<functional>`` if you must
```
jdswinbank: s/preffer/prefer/

PR #40, opened 2016-06-08
Add developer guide for eigen in coding/using_eigen.rst

timj:


timj:
@ziggyman you need to add this page to the table of contents before it can build properly.

coding/using_eigen.rst
```diff
@@ -0,0 +1,11 @@
+###########
+Using Eigen
+###########
+
+.. _cpp_using_eigen:
+
+Eigen offers some **unsupported** modules which are located in the ``include/unsupported`` directory. For details, refer to the `relevant Eigen documentation`_.
```
timj: I'd like an opening sentence stating that people can use any part of eigen that is labeled supported before we talk about what is forbidden. This guide needs to tell people what they can do before telling them what they can't.

PR #43, opened 2016-06-24
Add lfs batch config lines

jonathansick:
The purpose of this change is to repeat the result from a git config --global lfs.batch false to remind folks that this configuration is needed in case the earlier documentation lines were skipped. See
https://community.lsst.org/t/issues-using-git-lfs-for-cloning-test-data-repositories/880/16?u=jsick

https://developer.lsst.io/v/u-jonathansick-git-lfs-doc/tools/git_lfs.html

### DIFF
--- a/tools/git_lfs.rst
+++ b/tools/git_lfs.rst
@@ -219,7 +219,7 @@ In the repository's :file:`README`, we recommend that you include this section:
 
    To clone and use this repository, you'll need Git Large File Storage (LFS).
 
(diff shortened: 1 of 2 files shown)


jonathansick:
@jmatt See diff and link above. Does this look good to you?

jmatt:
@jsick Looks good. Exactly what we talked about.

jmatt:
Not related to this but related to using https now. L222 of git_lfs.rst is still using http instead of https. It forward just fine though.

jonathansick:
Right, will fix.

PR #46, opened 2016-07-26
Revise management including new TCT

ktlim:
This is intended to replace Frossie's tickets/DM-5888 branch.

### DIFF
--- a/processes/decision_process.rst
+++ b/processes/decision_process.rst
@@ -15,7 +15,7 @@ The :ref:`Request for Discussion (RFD) <decision-making-rfd>` process facilitate
 Empowerment
 ===========
 
-You are empowered by the DM Project Manager (Jeff Kantor), Project Scientist (Mario Juric) and Project Engineer (Kian-Tat Lim) to make decisions on any DM-internal matter---such as technical/algorithm issues, process improvements, and tool choices---when:
+You are empowered by the DM Subsystem Lead to make decisions on any DM-internal matter---such as technical/algorithm issues, process improvements, and tool choices---when:
 
 1. you are willing and able to do the work to implement the decision yourself or with people who agree with you,
 2. you (collectively) are willing and able to fix any problems if it goes wrong, and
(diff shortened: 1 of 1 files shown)


mjuric:
@ktlim, thanks!

One global proposal: strike references to names (i.e. Zeljko Ivezic), leave only roles (i.e., "DM Project Scientist"). Updating names when the roles change will be quite time consuming (and there's a danger we don't get to it).

PS: In the long-run, I'm also sure @jonathansick could come up with some clever and cool name-resolution mechanism for roles->names (`dig -t txt projectscientist.dm.lsst.org`?)  :)

processes/decision_process.rst
```diff
@@ -128,9 +127,9 @@ At least one member of the DM TCT_ will read each RFC to determine if it might a
 Responsibility and delegation
 -----------------------------
 
-For project management purposes, RFCs are formally proposals made to the Project Manager, Project Scientist and Project Engineer who by default are responsible for everything in DM (they "own" all problems).
-As owners, they have the final word in accepting or rejecting all proposals.
-Functionally, they delegate that ownership---the right and responsibility to make decisions---to others within the team (e.g. the System Architect, Interface Scientist, group leads, etc.) who are expected to delegate it even further.
+For project management purposes, RFCs are formally proposals made to the Subsystem Lead who by default is responsible for everything in DM (the Subsystem Lead "owns" all problems).
```
mjuric: This feels wrong as it skips many levels of the hierarchy. But I don't have anything better to replace it with.

jonathansick:
One thing we could do is store real names for position holders in variables so that in the future, changes only need to be made in one place. (The variables could be in the Sphinx conf.py config file, or even a yaml/json file in the repo).

Along the same lines of thinking, is there any opinion on whether it's time to move some of the DM organizational documents currently in Confluence (like https://confluence.lsstcorp.org/display/DM/Technical+Control+Team) into the developer.lsst.io?

jbecla:
Keep in mind that not only names but the whole structure might change as we switch from interim to permanent structure.

PR #52, opened 2016-08-08
DM-7150: Update Python testing section with log initialization info

hchiang2:


coding/python_testing.rst
```diff
@@ -208,6 +208,16 @@ final test suite.
 For the memory test to function properly the :lfunc:`lsst.utils.tests.init()` function must be invoked before any of the tests in the class are executed.
 Since LSST test scripts are required to run properly from the command-line and when called from within `pytest`_, the :lfunc:`~lsst.utils.tests.init()` function has to be in the file twice: once in the :ref:`setup_module <pytest:xunitsetup>` function that is called by `pytest`_ whenever a test module is loaded (`pytest`_ will not use the ``__main__`` code path), and also just before the call to :func:`unittest.main()` call to handle being called with :command:`python`.
 
+Utility initialization
+---------------------------------------
+:lfunc:`lsst.utils.tests.init()` does two things:
+initialize the memory tester as described previously, and
+initialize ``lsst.log`` by setting up a default configuration that
+outputs logs to console at ``lsst.log.INFO`` level.
+If you use ``lsst.log`` during tests, you may want to call
```
timj: `:lmod:` here

coding/python_testing.rst
```diff
@@ -208,6 +208,16 @@ final test suite.
 For the memory test to function properly the :lfunc:`lsst.utils.tests.init()` function must be invoked before any of the tests in the class are executed.
 Since LSST test scripts are required to run properly from the command-line and when called from within `pytest`_, the :lfunc:`~lsst.utils.tests.init()` function has to be in the file twice: once in the :ref:`setup_module <pytest:xunitsetup>` function that is called by `pytest`_ whenever a test module is loaded (`pytest`_ will not use the ``__main__`` code path), and also just before the call to :func:`unittest.main()` call to handle being called with :command:`python`.
 
+Utility initialization
+---------------------------------------
+:lfunc:`lsst.utils.tests.init()` does two things:
+initialize the memory tester as described previously, and
+initialize ``lsst.log`` by setting up a default configuration that
```
timj: `:lmod:`

coding/python_testing.rst
```diff
@@ -208,6 +208,16 @@ final test suite.
...
+outputs logs to console at ``lsst.log.INFO`` level.
```
timj: @jonathansick In the future how would we make that link to the `INFO` docs?

jonathansick:
Hi @hchiang2, thanks for this.

Could you shorten the header underline for Utilitiy initialization to match the length of the title text>

And also maybe switch to one-sentence per line line wrapping, like the rest of the document?

jonathansick:
Also leave a blank line between the headline/underline and the paragraph below it. Thanks!

coding/unit_test_snippets/testExecutables.py
```diff
@@ -9,4 +9,5 @@ class UtilsBinaryTester(lsst.utils.tests.ExecutablesTestCase):
 UtilsBinaryTester.create_executable_tests(__file__, EXECUTABLES)
 
 if __name__ == "__main__":
+    lsst.utils.tests.init()
```
timj: Do we need this? This test script runs executable C++ binaries so I'm assuming it won't inherit the logger setting from the parent python? Or does it? If we do need it then we will also need a `setup_module` in here so that `py.test` will work.

coding/python_testing.rst
timj: If we used `:lfunc:` I guess that would not work because it's a class attribute. Is there an `:lvar:` or `:lattr:` variant?

coding/unit_test_snippets/testExecutables.py
hchiang2: Removed.

hchiang2:
The corrections have been done -- thanks a lot for the help. 

I didn't link the class attribute. What is the correct way?

jonathansick:
`lsst.log` is a class?

timj:
`lsst.log.INFO` is a package global constant...

jonathansick:
In that case, use

```
:ldata:`lsst.log.INFO`
```

(see http://www.sphinx-doc.org/en/stable/domains.html#directive-py:data)

For class attributes it would have been

```
:lattr:`lsst.module.Class.attribute`
```

PR #53, opened 2016-08-15
DM-6823: Update test boilerplate and update testing intro

parejkoj:
Create pull request for comments.


coding/python_testing.rst
```diff
@@ -183,7 +183,7 @@ including :lclass:`~lsst.utils.tests.MemoryTestCase`:
 .. literalinclude:: unit_test_snippets/unittest_runner_example.py
    :linenos:
    :language: python
-   :emphasize-lines: 6, 16, 21
+   :emphasize-lines: 12, 17, 21
```
parejkoj: Why the emphasize lines here? It's not clear why these particular lines are marked. You should either have some text calling out why those are marked, or not mark them.

timj: They are the lines relevant to memory leak testing. This is the memory testing section.

timj: I'll add a comment about that to the text.

PR #56, opened 2016-08-25
DM-5456: Python Style Guide as a diff of PEP 8 and accept Numpydoc

jonathansick:
This PR updates the Python Style Guide in response to two accepted RFCs:

- [RFC 162: LSST Python Style Guide delta to PEP 8](http://ls.st/rfc-162)
- [RFC 214: Use numpydoc docstrings](http://ls.st/rfc-214)

Both the 'Python Style Guide' and 'Documenting Python APIs' pages are being updated in the same ticket since they effectively form the Python Style Guide as a whole—the numpydoc material is simply too detailed to include with the rest of the style guide.

Live preview of these revised pages:

- https://developer.lsst.io/v/DM-5456/coding/python_style_guide.html
- https://developer.lsst.io/v/DM-5456/docs/py_docs.html

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -251,148 +251,50 @@ Other subtleties have been noted in https://fuhm.net/super-harmful/:
 
 .. _style-guide-py-files:
 
-4. Files
-========
+4. Source Files & Modules
+=========================
 
-.. _style-guide-py-4-1:
+.. _style-guide-py-file-name:
 
-A Python source file name SHOULD be camelCase-with-leading-lowercase and ending in ``.py``
-------------------------------------------------------------------------------------------
+A Python source file name SHOULD be camelCase-with-leading-lowercase and ending in '.py'
+----------------------------------------------------------------------------------------
 
-The name of a file containing a module will be the ``camelCase``-with-leading-lowercase transliteration of the module name.
+A module containing a single class should be a ``camelCase``-with-leading-lowercase transliteration of the class's name.
+
 The name of a test case should be descriptive without the need for a trailing numeral to distinguish one test case from another. 
 
-.. _style-guide-py-4-2:
+.. TODO consider refactoring tests into their own section
+
+.. _style-guide-py-file-encoding:
 
 ASCII Encoding MUST be used for new code
 ----------------------------------------
 
-- Always use ASCII for new python code.
+Always use ASCII for new python code.
 
 - **Do not** include a coding comment (as described in  :pep:`263`) for ASCII files.
 
 - Existing code using Latin-1 encoding (a.k.a. ISO-8859-1) is acceptable so long as it has a proper coding comment. All other code must be converted to ASCII or Latin-1 except for 3rd party packages used "as is."
 
-.. _style-guide-py-4-3:
+.. _style-guide-py-file-order:
 
-DM specified code order SHOULD be followed
-------------------------------------------
+Standard code order SHOULD be followed
+--------------------------------------
 
 Within a module, follow the order: 
 
-1. Shebang line (``#!``), only for executable scripts
-2. Module-level comments
+1. Shebang line, ``#! /usr/bin/env python`` (only for executable scripts)
+2. Module-level comments (such as the `license statement <https://github.com/lsst/templates/blob/master/CopyrightHeader.py>`__)
 3. Module-level docstring
 4. Imports
 5. ``__all__`` statement, if any
-6. Module variables (names start with underscore)
-7. Module functions and classes (names start with underscore)
-8. Public variables
+6. Private module variables (names start with underscore)
+7. Private module functions and classes (names start with underscore)
+8. Public module variables
 9. Public functions and classes
 10. Optional test suites
 
-.. _style-guide-py-4-4:
-
-Imports SHOULD be grouped, in order: standard lib, 3rd party lib, local lib
----------------------------------------------------------------------------
-
-Imports should be grouped in the following order, with each group separated by a blank line:
-
-1. standard library imports
-2. related third party imports
-3. local application/library specific imports
-
-When importing a class from a class-containing module,
-it's usually okay to do this:
-
-.. code-block:: py
-
-   from myclass import MyClass
-   from foo.bar.yourclass import YourClass
-
-But if that causes local name clashes, then do this instead: 
-
-.. code-block:: py
-
-   import myclass
-   import foo.bar.yourclass
-
-and use ``myclass.MyClass`` and ``foo.bar.yourclass.YourClass``. 
-
-Relative imports SHOULD be used when importing another module from the same package
-Consider this layout: 
-
-.. code-block:: text
-
-   mypkg/
-       __init__.py
-       foo.py
-       bar.py
-
-If ``foo`` wants to import ``bar``, the safe way to do this (Python 2.6 and later) is to use relative import:
-
-.. code-block:: py
-
-   from . import bar
-
-Or, if you just want a few symbols from bar, this also works:
-
-.. code-block:: py
-
-   from .bar import thing1, thing2
-
-This avoids any danger of name collision with a module on the python path named bar.
-Relative import statements are richer than suggested by the example; see :pep:`328` for details.
-
-.. _style-guide-py-4-5:
-
-Multiple Statements SHOULD NOT occur on same Line
--------------------------------------------------
-
-Compound statements (multiple statements on the same line) are generally discouraged.
-
-Yes: 
-
-.. code-block:: py
-
-   if foo == 'blah':
-       do_blah_thing()
-       do_one()
-       do_two()
-       do_three()
-
-No:
-
-.. code-block:: py
-
-   if foo == 'blah': do_blah_thing()
-       do_one(); do_two(); do_three()
-
-While sometimes it's okay to put an ``if``/``for``/``while`` with a small body on the same line, never do this for multi-clause statements.
-Also avoid folding such long lines!
-
-Rather not:
-
-.. code-block:: py
-
-   if foo == 'blah': do_blah_thing()
-   for x in lst: total += x
-   while t < 10: t = delay()
-
-Definitely not:
-
-.. code-block:: py
-
-   if foo == 'blah': do_blah_thing()
-   else: do_non_blah_thing()
- 
-   try: something()
-   finally: cleanup()
- 
-   do_one(); do_two(); do_three(long, argument,
-                                list, like, this)
-                            
-   if foo == 'blah': one(); two(); three()
+.. note:: Delete mention of test suites?
 
 .. _style-guide-py-string-handling:
 
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -149,455 +149,10 @@ Style changes must be encapsulated in a distinct commit (see :ref:`git-commit-or
 
 .. _autopep8: https://pypi.python.org/pypi/autopep8
 
-.. _style-guide-py-naming:
-
-2. Naming Conventions
-=====================
-
-We follow `PEP 8ʼs naming conventions <https://www.python.org/dev/peps/pep-0008/#naming-conventions>`_, with exceptions listed here.
-The naming conventions for LSST Python and C++ source have been defined to be as similar as the respective languages allow.
-
-In general:
-
-- class names are ``CamelCase`` with leading uppercase,
-- module variables used as module global constants are ``UPPERCASE_WITH_UNDERSCORES``,
-- all other names are ``camelCase`` with leading lowercase.
-
-Names may be decorated with leading and/or trailing underscores.
-
-.. _style-guide-py-2-2:
-
-User defined names SHOULD NOT shadow python built-in functions
---------------------------------------------------------------
-
-Names which shadow a python built-in function may cause confusion for readers of the code.
-Creating a more specific identifier is suggested to avoid collisions.
-In the case of *filter*, ``filterName`` may be appropriate; for *filter objects*, something like ``filterObj`` might be appropriate.
-
-.. _style-guide-py-naming-attributes:
-
-Class Attribute Names SHOULD be camelCase with leading lowercase
-----------------------------------------------------------------
-
-`Opposes PEP 8 <https://www.python.org/dev/peps/pep-0008/#id45>`__.
-Error codes: N802 and N803.
-
-.. _style-guide-py-naming-functions:
-
-Module methods (free functions) SHOULD be camelCase with leading lowercase
---------------------------------------------------------------------------
-
-`Opposes PEP 8 <https://www.python.org/dev/peps/pep-0008/#id45>`__.
-Error code: N802.
-
-.. _style-guide-py-naming-class-modules:
-
-Modules which contain class definitions SHOULD be named after the class name
-----------------------------------------------------------------------------
-
-Modules which contain class definitions should be named after the class name (one module per class).
-
-.. _style-guide-py-naming-ext-modules:
-
-When a Python module wraps a C/C++ extension module, the C/C++ module SHOULD be named <module>Lib
--------------------------------------------------------------------------------------------------
-
-When an extension module written in C or C++ has an accompanying Python module that provides a higher level (e.g. more object oriented) interface, the C/C++ module should append ``Lib`` to the module's name (e.g. ``socketLib``).
-
-.. _style-guide-py-naming-ambiguous:
-
-Names l (lowercase: el), O (uppercase: oh), I (uppercase: eye) MUST be avoided
-------------------------------------------------------------------------------
-
-Never use these characters as single character variable names:
-
-- ``l`` (lowercase letter el),
-- ``O`` (uppercase letter oh), or
-- ``I`` (uppercase letter eye).
-
-In some fonts, these characters are indistinguishable from the numerals one and zero.
-When tempted to use ``l``, use ``L`` instead.
-
-.. _style-guide-py-inheritance:
-
-3. Designing for Inheritance
-============================
-
-Always decide whether a class's methods and instance variables (collectively: "attributes") should be public or non-public.
-If in doubt, choose non-public; it's easier to make it public later than to make a public attribute non-public.
-
-Public attributes are those that you expect unrelated clients of your class to use, with your commitment to avoid backward incompatible changes.
-Non-public attributes are those that are not intended to be used by third parties; you make no guarantees that non-public attributes won't change or even be removed.
-
-We don't use the term "private" here, since no attribute is really private in Python (without a generally unnecessary amount of work).
-Another category of attributes are those that are part of the "subclass API" (often called "protected" in other languages).
-Some classes are designed to be inherited from, either to extend or modify aspects of the class's behavior.
-When designing such a class, take care to make explicit decisions about which attributes are public, which are part of the subclass API, and which are truly only to be used by your base class.
-
-For simple public data attributes, it is best to expose just the attribute name, without complicated accessor/mutator methods.
-Keep in mind that Python provides an easy path to future enhancement, should you find that a simple data attribute needs to grow functional behavior.
-In that case, use properties to hide functional implementation behind simple data attribute access syntax.
-
-- Note 1: Properties only work on new-style classes.
-
-- Note 2: Try to keep the functional behavior side-effect free, although side-effects such as caching are generally fine.
-
-- Note 3: Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.
-
-.. _style-guide-py-super:
-
-``super`` SHOULD NOT be used unless the author really understands the implications (e.g. in a well-understood multiple inheritance hierarchy).
-----------------------------------------------------------------------------------------------------------------------------------------------
-
-Python provides ``super`` so that each parent class' method is only called once (see https://www.python.org/download/releases/2.3/mro/).
-The problem is, if you're going to use super at all, then all parent classes in the chain (also called the Method Resolution Order") need to use super otherwise the chain gets interrupted. 
-Other subtleties have been noted in https://fuhm.net/super-harmful/:
-
-- Never call super with anything but the exact arguments you received, unless you really know what you're doing.
-- When you use it on methods whose acceptable arguments can be altered on a subclass via addition of more optional arguments, always accept ``*args, **kw``, and call ``super`` like ``super(MyClass, self).currentmethod(alltheargsideclared, *args, **kwargs)``.
-  If you don't do this, forbid addition of optional arguments in subclasses.
-- Never use positional arguments in ``__init__`` or ``__new__``.
-  Always use keyword args, and always call them as keywords, and always pass all keywords on to ``super``.
-
-.. _style-guide-py-files:
-
-4. Source Files & Modules
-=========================
-
-.. _style-guide-py-file-name:
-
-A Python source file name SHOULD be camelCase-with-leading-lowercase and ending in '.py'
-----------------------------------------------------------------------------------------
-
-A module containing a single class should be a ``camelCase``-with-leading-lowercase transliteration of the class's name.
-
-The name of a test case should be descriptive without the need for a trailing numeral to distinguish one test case from another. 
-
-.. TODO consider refactoring tests into their own section
-
-.. _style-guide-py-file-encoding:
-
-ASCII Encoding MUST be used for new code
-----------------------------------------
-
-Always use ASCII for new python code.
-
-- **Do not** include a coding comment (as described in  :pep:`263`) for ASCII files.
-
-- Existing code using Latin-1 encoding (a.k.a. ISO-8859-1) is acceptable so long as it has a proper coding comment. All other code must be converted to ASCII or Latin-1 except for 3rd party packages used "as is."
-
-.. _style-guide-py-file-order:
-
-Standard code order SHOULD be followed
---------------------------------------
-
-Within a module, follow the order: 
-
-1. Shebang line, ``#! /usr/bin/env python`` (only for executable scripts)
-2. Module-level comments (such as the `license statement <https://github.com/lsst/templates/blob/master/CopyrightHeader.py>`__)
-3. Module-level docstring
-4. Imports
-5. ``__all__`` statement, if any
-6. Private module variables (names start with underscore)
-7. Private module functions and classes (names start with underscore)
-8. Public module variables
-9. Public functions and classes
-10. Optional test suites
-
-.. note:: Delete mention of test suites?
-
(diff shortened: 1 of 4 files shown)


timj:
I love how short the first section is now.

For the `noqa` item, please also state that the comment should include the E code that is being hidden. `# noqa` on its own is always confusing later when someone looks at the code and tries to work out why it's been hidden.

When we list the PEP8 exceptions, we should include the justification text from the previous coding standard. Otherwise people reading the new version can't work out why we are ignoring them.

I keep forgetting that [RFC-97](https://jira.lsstcorp.org/browse/RFC-97) is still in a flagged state pending approval of example text appearing for review. This version still has the old wording.

The remaining comments are more for the record, rather than having to be fixed in this PR.

Looking at the style guide as if it is a unified entity rather than looking at the diff, I'd like to suggest we take the opportunity to remove the entire section on `apply()`. This function was deprecated in Python 2.3 around 2004  and does not exist on Python 3. It's existence in the style guide simply tells us that the style guide was started a long time ago and many people won't even know what the `apply()` function was for. I will have to remove it as part of the Python 3 tweak that I have to make.

That section on python 2.5 exceptions will be disappearing soon as well because Python 3 won't let you use that style anyhow.

There is a rST error in the `set()` section.

jonathansick:
Thanks @timj, I'll work on

- [ ] Adding justification text for PEP 8 exception codes
- [ ] Fixing the `set()` section
- [ ] Adding that `noqa` comments should state the targeted error code

Some other points:

## Re: RFC-97

If I can help move [RFC-97](https://jira.lsstcorp.org/browse/RFC-97) along I'm happy to do so. In RFC-97 @ktlim seemed to mention two possible policies for new code

> The simplest route to adoption is to say that it applies to independent top-level products (e.g. those built by Jenkins, or independent services such as those provided by SQuaRE) including all their first-party (LSST-developed) dependencies; each such product is either all PEP8-named or all camelCase-named. I think that several products have already gone this route. The problem with this is that it discourages sharing code.

vs.

> An alternative would be to say that it applies to new packages (or git repositories), which would have the option of being PEP8-named (in Python and C++). I think this has higher costs in terms of cognitive load and transition impacts.

If I can get some guidance/go-ahead I'll put that text in this PR.

## Organization

Another issue I'm thinking about is whether the style guide might benefit from a general top-level re-organization. Currently the sections are:

1. PEP 8 is the Baseline Coding Style
2. Naming Conventions
3. Designing for Inheritance
4. Source File & Modules
5. Comparisons
6. Programming Pitfalls
7. Programming Recommendations
8. Suggested Modules
9. Python 3 Idioms
10. Layout
11. Whitespace
12. Comments
14. Documentation Strings (docstrings)

'Designing for Inheritance' might be better rephrased as 'Classes.' The intro to the 'Designing for Inheritance' section actually makes a number of recommendations that might be better phrased as rules (public vs private APIs; opting for simple attributes first).

'Programming Pitfalls' and 'Programming Recommendations' seem to be an unclear distinction (and have also been hollowed-out with this refactoring. Given that Python 3 is now *a thing for us*, it might be better to combine those sections with 'Python 3 idioms' into a new 'Idioms' section.

In addition, I think there could be useful to re-order the sections for better flow.

So here's my proposal, ordered from general to specific:

1. PEP 8 is the Baseline Coding Style
2. Layout
3. Whitespace
4. Comments
5. Documentation Strings (docstrings)
6. Naming Conventions
7. Source File & Modules
8. Classes *(formerly Designing for Inheritance)*
9. Comparisons
10. Idioms *(formerly Programming Pitfalls, Programming Recommendations and Python 3 Idioms)*

timj:
I like the proposed new layout. Sections 1 to 7 are clearly part of the style guide. The remaining sections seem to be programming advice.

Regarding RFC-97.  I think the two paragraphs after the two you quoted are relevant:

> For either level, I'm not sure I really like the "allowed" and "encouraged" language. I think it's clearer for developers if we pick a stance and stick with it, even if we have (temporary) non-conformance. So I would be a bit happier if this were to say that all new products or packages must be PEP8-named.

A clear statement here about what to do with new packages. They will be PEP8 compliant. End of discussion. I think that this can be added directly to the naming section.

> Transitioning existing code to PEP8-naming is a bit of a bigger deal, since it involves code in not just the package transitioning but also all packages dependent on it. It would help if good tools could be made available to assist this process.

This paragraph is dealing with the existing code base. It ties into the whole debate over [RFC-81](https://jira.lsstcorp.org/browse/RFC-81) because it raises the possibility of changes to the C++ code. I think a transition in APIs to PEP8 naming will have to be considered as part of that RFC (with a shim layer providing backwards compatibility with deprecation warnings). What we can tackle is the debate over local variables. I think if we can have a plan for local variables (that are not part of the API and just used in a single file) then that would be a good place to begin. We can make a transition like that on a file by file basis without affecting any other part of the code. We can just do it when we are working on a file for some other reason as part of a `flake8` pass.

parejkoj:
I like all of @timj 's comments. I'll take another look once you've reorganized it.

jonathansick:
@timj and @parejkoj I've updated https://developer.lsst.io/v/DM-5456/coding/python_style_guide.html based on the recommendations and with the re-organization.

Do you think I should put PEP 8 naming text with this ticket, or is that over-reach?

timj:
I think the naming is over-reach for this ticket but we should jump on it as soon as @ktlim gets back from vacation. I'm pretty sure I've interpreted K-T's meaning from the RFC but we should check. I should be able to approve this PR on my own because it's not technically changing any policy. Then I think the naming RFC would go a lot faster if there was a branch made for that that K-T could read.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
-.. _style-guide-py-2-3:
+Linters are installable with :command:`pip`:
 
-Class names SHOULD use CamelCase with leading uppercase
--------------------------------------------------------
+.. code-block:: bash
 
-Python class names should follow the same conventions as C++ class names---they should be ``CamelCase`` with leading uppercase.
-Note that exceptions are classes and thus follow the same convention.
+   pip install flake8
```
timj:
```
conda install flake8
```
works as well.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
 
-Exception names relating to errors SHOULD include the suffix Error
-------------------------------------------------------------------
+.. code-block:: bash
 
-An error exception name SHOULD include the suffix ``Error``.
+   flake8 --ignore=E133,E226,E228,N802,N803 --max-line-length=110 {{python_dir}}
 
-.. _style-guide-py-2-5:
+where ``{{python_dir}}`` is a directory with Python source files.
```
timj: You can give `.` as the directory and `flake8` will then go through all the subdirectories so it doesn't have to be "a directory with Python source files" (which implies only the python files in the specified directory).

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
+.. code-block:: ini
 
-Module methods (free functions) SHOULD be camelCase with leading lowercase
---------------------------------------------------------------------------
+	[flake8]
+	max-line-length = 110
+	ignore = E133, E226, E228, E251, N802, N803
 
----
+:command:`flake8` can be invoked without arguments when this configuration is present.
```
timj: The trick is the inconsistent behavior that this can lead to if people forget which packages have them and which don't.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
-.. _style-guide-py-2-4:
+Flake8 command line invocation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Exception names relating to errors SHOULD include the suffix Error
-------------------------------------------------------------------
+.. code-block:: bash
 
-An error exception name SHOULD include the suffix ``Error``.
+   flake8 --ignore=E133,E226,E228,N802,N803 --max-line-length=110 {{python_dir}}
```
timj: we should mention at this point that the specific error code meanings will be discussed later on. Otherwise you immediately have a question about what E133 means. The alternative is to bring the PEP8 exceptions list forward so it appears before the `flake8` subsection. You then provide the exceptions and then say how you can check for them.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
 
-For simple public data attributes, it is best to expose just the attribute name, without complicated accessor/mutator methods.
-Keep in mind that Python provides an easy path to future enhancement, should you find that a simple data attribute needs to grow functional behavior.
-In that case, use properties to hide functional implementation behind simple data attribute access syntax.
+    class Rectangle(Blob):
+        """Documentation for Rectangle.
+        """
+        def __init__(self, width, height,
+                     color='black', emphasis=None, highlight=0):
+            if width == 0 and height == 0 and
```
timj: This example code is syntactically broken so I'm not sure what to make of it. It's broken because it doesn't do continuation properly but I think it's point should be that the first `if` statement uses `\` for continuation (those are missing) and the second one uses parens. It might be worth noting that:
```python
if (width == 0
    height == 0):
    do_something()
```
can be visually confusing with the last line of the `if` lining up with the statement inside the block.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
 
-If your class is intended to be subclassed, and you have attributes that you do not want subclasses to use, name them with double leading underscores and no trailing underscores.
-This invokes Python's name mangling algorithm, where the name of the class is mangled into the attribute name.
-This helps avoid attribute name collisions should subclasses inadvertently contain attributes with the same name.
+.. seealso::
 
-- Only the simple class name is used in the mangled name, so if a subclass chooses both the same class name and attribute name, you can still get name collisions.
+   - `flake8 error codes <https://flake8.readthedocs.io/en/latest/user/error-codes.html>`_
+   - `pycodestyle error codes <https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes>`_
+   - `pep8-naming error codes <https://github.com/PyCQA/pep8-naming#plugin-for-flake8>`_
```
timj: Excellent.

coding/python_style_guide.rst
timj: Actually, make clear you can also specify a single file.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
-Always decide whether a class's methods and instance variables (collectively: "attributes") should be public or non-public.
-If in doubt, choose non-public; it's easier to make it public later than to make a public attribute non-public.
+Python's implied continuation inside parens, brackets and braces SHOULD be used for wrapped lines
+-------------------------------------------------------------------------------------------------
 
-Public attributes are those that you expect unrelated clients of your class to use, with your commitment to avoid backward incompatible changes.
-Non-public attributes are those that are not intended to be used by third parties; you make no guarantees that non-public attributes won't change or even be removed.
+The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces.
+If necessary, you can add an extra pair of parentheses around an expression, but sometimes using a backslash looks better.
+Make sure to indent the continued line appropriately. Some examples:
```
timj: I have a slight quibble over "Some examples" vs "An example" but I suppose it depends on whether you think there is a single code example here or many examples in one snippet.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
-.. code-block:: py
+Never surround these binary arithmetic operators with whitespace:
 
-   from myclass import MyClass
-   from foo.bar.yourclass import YourClass
+- multiplication (``*``),
+- division (``/``),
+- exponentiation (``**``),
+- floor division (``//``),
+- modulus (``%``).
```
timj: I believe that we do want spaces around `%` when it is used for string formatting (as opposed to as a math operator).

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
-   import foo.bar.yourclass
+   i = i + 1
+   submitted += 1
+   x = x*2 - 1
+   hypot2 = x*x + y*y
+   c = (a + b)*(a - b)
 
-and use ``myclass.MyClass`` and ``foo.bar.yourclass.YourClass``. 
+This deviates from PEP 8, which `allows whitespace around these arithmetic operators if they appear alone <https://www.python.org/dev/peps/pep-0008/#id28>`__.
+Error codes: N226 and N228.
```
timj: `E` not `N`.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
+       karg3=value3,
+   )
 
-This avoids any danger of name collision with a module on the python path named bar.
-Relative import statements are richer than suggested by the example; see :pep:`328` for details.
+   aFunction(x, y, z, karg1 = value1, karg2 = value2, karg3 = value3)
 
-.. _style-guide-py-4-5:
+`Opposes PEP 8 <https://www.python.org/dev/peps/pep-0008/#id28>`__.
+Error code: N251.
```
timj: This implies the autopep8 line above should include `N251`. Leave it out of the `flake8` line because it half does the right thing.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
 
-5. String Handling
-==================
+5. Documentation Strings (docstrings)
+=====================================
 
-.. _style-guide-py-string-5-1:
+Use **Numpydoc** to format the content of all docstrings.
+The page :doc:`../docs/py_docs` authoritatively describes this format.
+It's guide should be treated as an extension of this Python style guide.
```
timj: "Its"

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
+------------------------------------------------------------------------------------
 
-``.startswith()`` and ``.endswith()`` SHOULD be used to check for prefixes or suffixes
---------------------------------------------------------------------------------------
+Write docstrings for all public modules, functions, classes, and methods.
+See :doc:`../docs/py_docs`.
 
-Use :py:meth:`str.startswith()` and :py:meth:`str.endswith()` instead of string slicing to check for prefixes or suffixes; they are cleaner and less error prone.
+Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does.
+This comment should appear after the ``def`` line.
```
timj: What's the downside of that comment just being a docstring?

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
+6. Private module variables (names start with underscore)
+7. Private module functions and classes (names start with underscore)
+8. Public module variables
+9. Public functions and classes
+10. Optional test suites
 
-   def proclist(alist=None):
-   if alist == None:
-   alist = []
+.. note:: Delete mention of test suites?
```
timj: Depends whether the word "suite" means `TestSuite` or if it means generically some test code. I think we need some clarification on what that means before deleting it so I think we have to keep it in for the moment and flag it with @ktlim for later removal.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
 
-Object type comparisons SHOULD always use ``isinstance()``
-----------------------------------------------------------
+.. _style-guide-py-super:
 
-Object type comparisons should always use :py:func:`isinstance()` instead of comparing types directly. 
+``super`` SHOULD NOT be used unless the author really understands the implications (e.g. in a well-understood multiple inheritance hierarchy).
+----------------------------------------------------------------------------------------------------------------------------------------------
+
+Python provides ``super`` so that each parent class' method is only called once (see https://www.python.org/download/releases/2.3/mro/).
```
timj: I wonder if a newer document could be found.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
-.. _style-guide-py-8-3:
+.. _style-guide-py-open:
 
-``raise ValueError('message')`` SHOULD be used instead of the deprecated form
------------------------------------------------------------------------------
+Use ``open`` instead of ``file``
+--------------------------------
 
-When raising an exception, use ``raise ValueError('message')`` instead of the older, deprecated form ``raise ValueError, 'message'``.
+This is preferred and ``file`` is gone in Python 3.
```
timj: I think I'd just drop the "This is preferred and" prefix because we not giving people a choice.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
 
-.. _style-guide-py-suggested-modules:
+.. note:: Should the :py:mod:`io` module be recommended more specifically?
```
timj: I think `io` is a distraction in this section. `open` is available regardless.

coding/python_style_guide.rst
```diff
@@ -12,648 +13,671 @@ To request changes to these standards, please file an :ref:`RFC <decision-making
...
+.. _style-guide-py-exception-handling-syntax:
 
-.. _style-guide-py-9-1:
+.. _style-guide-py-subprocess:
 
 The ``subprocess`` module SHOULD be used to spawn processes
 -----------------------------------------------------------
 
 Use the :py:mod:`subprocess` module to spawn processes.
 This supersedes and unifies :py:func:`os.system`, ``os.spawn``, :py:func:`os.popen`, etc..
```
timj: `os.spawn` is so old we can't even link to the documentation for it...

coding/python_style_guide.rst
```diff
@@ -662,25 +686,39 @@ This means ``/`` is floating-point division and ``//`` is truncated integer divi
...
+
+Use ``as`` when catching an exception
+-------------------------------------
+
+For example, use ``except Exception as e`` or ``except (LookupError, TypeError) as e``.
+The new syntax is clearer, especially when catching multiple exception classes, and required in Python 3.
+
+.. _style-guide-py-generators:
+
+Iterators and generators SHOULD be used to iterate overlarge data sets efficiently
```
timj: I think this should be "over large"

coding/python_style_guide.rst
```diff
@@ -662,25 +686,39 @@ This means ``/`` is floating-point division and ``//`` is truncated integer divi
...
+.. _style-guide-py-future-itervalues:
 
 Use ``itervalues()`` and ``iteritems()`` instead of ``values()`` and ``items()``
 --------------------------------------------------------------------------------
 
 For iterating over dictionary values and items use the above idiom unless you truly need a list.
-This generates more efficient code today and helps 2to3_ generate more efficient code in the future.
-For more information see http://python3porting.com/preparing.html#optional-use-the-iterator-methods-on-dictionaries.
+This generates more efficient code today and helps futurize_ generate more efficient code in the future.
+For more information see http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items.
```
timj: with the caveat "unless the code has already been ported to support Python 3". `meas_algorithms` and below can not use `itervalues` etc.

coding/python_style_guide.rst
```diff
@@ -701,429 +739,17 @@ To test for inclusion use ``in``:
...
-Delimit doc strings with ``"""`` (three double quotes). You may use ``u"""`` for unicode but it is usually preferable to stick to ASCII.
-The terminating """ should be on its own line, even for one-line doc strings (this is a minor departure from :pep:`257`).
-
-.. code-block:: py
-
-   """Return a foobang
-    
-   Optional plotz says to frobnicate the bizbaz first.
-   """
+The special method ``next`` has been renamed to ``__next__`` in Python 3.
```
timj: and the `next()` function is supported on Python 2.7.

docs/py_docs.rst
```diff
@@ -21,102 +16,139 @@ We document Python code in three ways:
...
-This page focuses on public code documentation through docstrings, while the latter two are discussed in our Python style guide.
+This page focuses on public code documentation through docstrings, while the latter two are discussed in our :doc:`../coding/python_style_guide`.
 
-.. TODO add link to python style guide.
+Treat the guidelines on this page as an extension of the :doc:`../coding/python_style_guide`.
 
-.. _py-doc-boilerplate:
+.. note::
+   Changes to this document must be approved by the System Architect (`RFC-24 <https://jira.lsstcorp.org/browse/RFC-24>`_).
+   To request changes to these standards, please file an :ref:`RFC <decision-making-rfc>`.
```
timj: I think you need this note at the top of the Python style guide doc.

docs/py_docs.rst
```diff
@@ -528,7 +570,7 @@ E.g., whilst documenting a ``lsst.afw.tables`` module, refer to a class in ``lss
 
 .. code-block:: rst
 
-   afw.detection.Footprint : Regular detection footprint.
+   `afw.detection.Footprint` : Regular detection footprint.
```
timj: a few lines below this one it looks like the back ticks are missing from `astropy.table.Tables`.

coding/python_style_guide.rst
jonathansick: For implicity I'll ignore `conda install`—though maybe it'd be useful it flake8 et al came with the Stack?

coding/python_style_guide.rst
jonathansick: I'm not sure how to re-write this to make it better. Maybe we can do a ticket to merge a `setup.cfg` into all packages quickly so it's a non-issue.

coding/python_style_guide.rst
jonathansick: Good catch. I realize I was originally inconsistent with `N251` in my ignore lists, oops.

So now autoflake8 will ignore fixing `N251`, but flake8 *will* lint it. Let's flag `N251` for another RFC because that multi-line keyword arg assignment format is indefensible. :)

coding/python_style_guide.rst
jonathansick: No downside at all—I'm just preserving the original text. I'm in favour of docstrings everywhere, but I've gotten pushback from certain people when advocating it in code reviews...

Flag for another (minor) RFC?

coding/python_style_guide.rst
jonathansick: Surprisingly, not really. the 2.3 docs are the only ones with that page. https://docs.python.org/3/library/functions.html#super. But I suppose I could just link to Python 3's `super()` [docs](https://docs.python.org/3/library/functions.html#super) and Raymond Hettinger's [Super Considered Super!](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/) as a tonic to the blatant misinformation in https://fuhm.net/super-harmful/ that misled whoever wrote DM's original Python style guide.

parejkoj: Please, yes!

jonathansick:
I think I've covered all the comments. I need to give it another read-through to pick up remaining issues.

Also, I know that the alignment in https://developer.lsst.io/v/DM-5456/coding/python_style_guide.html#python-s-implied-continuation-inside-parens-brackets-and-braces-should-be-used-for-wrapped-lines is totally broken. I think it has to do with pseudo-bold being used in the code samples.

And to consolidate some discussion, these are changes that we can push to future tickets/PR discussions:

1. PEP 8 naming (RFC=97)
2. `super` being more acceptable
3. Elimination of non-PEP 8 whitespace rules in https://developer.lsst.io/v/DM-5456/coding/python_style_guide.html#keyword-assignment-operators-should-be-surrounded-by-a-space-when-statements-appear-on-multiple-lines

coding/python_style_guide.rst
```diff
@@ -2,658 +2,697 @@
...
 
-.. _style-guide-py-7-1:
+In some fonts, these characters are indistinguishable from the numerals one and zero.
+When tempted to use ``l``, use ``L`` instead.
 
-``if x`` SHOULD NOT be used when you mean ``if x != None``
-----------------------------------------------------------
+.. _style-guide-py-files:
 
-Beware of writing ``if x`` when you mean ``if x != None``.
```
timj: Did this disappear because PEP-8 says the same thing? If so, DM-6720 should be annotated to indicate that the work will be subsumed into this ticket.

jonathansick: See this section in PEP 8: https://www.python.org/dev/peps/pep-0008/#id49

Note how our original style guide actually made a mistake while lifting text from PEP 8:

> Also, beware of writing if x when you really mean if x is not None

I'll invalidate DM-6720 now.

timj: DM-6720 was entirely about that error in the text.

coding/python_style_guide.rst
```diff
@@ -2,658 +2,697 @@
...
+       karg3=value3,
+   )
 
-While sometimes it's okay to put an ``if``/``for``/``while`` with a small body on the same line, never do this for multi-clause statements.
-Also avoid folding such long lines!
+   aFunction(x, y, z, karg1 = value1, karg2 = value2, karg3 = value3)
 
-Rather not:
+`Opposes PEP 8 <https://www.python.org/dev/peps/pep-0008/#id28>`__.
+Error code: N251.
```
timj: E251

coding/python_style_guide.rst
```diff
@@ -2,658 +2,697 @@
...
-Failure to use the ``__all__`` mechanism results in all names in the module's namespace, which do not begin with a single ``_``, being exported as global.
+Many :pep:`8` issues in existing code can be fixed with `autopep8`_:
 
-.. _style-guide-py-2-13:
+.. code-block:: bash
 
-``self`` MUST be used for the first argument to instance methods
-----------------------------------------------------------------
+   autopep8 . --in-place --recursive \
+       --ignore E133,E226,E228,N251,N802,N803 --max-line-length 110
```
timj: E251.

jonathansick:
I think I've addressed all the in-scope comments. LGTM? Or should we try to quickly get the [Idioms](https://jira.lsstcorp.org/browse/DM-5456?focusedCommentId=53206&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-53206) section sorted?

PR #57, opened 2016-08-25
Quick spelling fix: baulk -> balk

mssgill:


jdswinbank:
Huh — apparently "baulk" is [British](http://www.oxfordlearnersdictionaries.com/us/definition/english/baulk). I suppose we have to lower ourselves to use the American variant in these docs! :-) Good catch.

mssgill:
Ah - i'm familiar with the normal UK variants, and wouldn't have piped up on those, but didn't note this one (and bet most folks wouldn't either)!  

Ok, the one other spelling prob in this doc is 'preceed' vs. 'precede' -- that's not a variant, right..?!

jdswinbank:
My dictionary assures me that "preceed" is "now nonstandard". :-)

PR #59, opened 2016-09-02
DM-6317: Update Python style guide for Python 3

timj:
* Python 3 now mandatory
* Python 2.7 optional
* Update Idioms section to remove mention of idioms that don't work on Python 3.

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -643,6 +643,40 @@ For more information see the online book `Supporting Python 3 <http://python3por
 
 .. _futurize: http://python-future.org/futurize.html
 
+Supporting Python 2.7 and 3.x simultaneously
+---------------------------------------------
+
+The ``future`` package MUST be used to provide compatibility functionality with Python 2.7
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We use the `future <http://python-future.org/>`_ package to provide a means for writing code using Python 3 idioms that will also work on Python 2.7.
+Details on the process for migrating a 2.7 codebase to support both versions can be found in `SQR-014 <https://sqr-014.lsst.io>`_.
+
+.. _style-guide-py-future-absolute-import:
+.. _style-guide-py-future-division:
+.. _style-guide-py-print:
+
+``__future__`` MUST be used to import Python 3 behavior in all files where the related functionality is used
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Code that is to be used on Python 2.7 and 3.x should import ``division``, ``print_function`` and ``absolute_import`` from the :mod:`__future__` package where appropriate.
+This means ``/`` is floating-point division and ``//`` is truncated integer division, regardless of the type of numbers being divided and matches the Python 3 behavior.
+
+In addition, import local modules using relative imports (e.g. ``from . import foo`` or ``from .foo import bar``).
+This results in clearer code and avoids shadowing global modules with local modules.
+
+The :py:func:`print()` function is required in Python 3.
+In general, DM code should use logging instead of ``print`` functions.
+
+
+.. _style-guide-py-future-itervalues:
+
+``itervalues()`` and ``iteritems()`` CANNOT be used
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Python 3 does not support the ``iter`` variants.
+For more information on how to handle this efficiently in Python 2 see http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items.
+
 .. _style-guide-py-pitfalls-mutables:
 
 A mutable object MUST NOT be used as a keyword argument default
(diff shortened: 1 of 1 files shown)


fred3m:
Currently the document is agnostic as to which version of python 3 should be supported. This might be difficult as we should be testing *all* versions of python 3 in Jenkins for each build, which can become cumbersome. I have seen instances in other codes where minor differences can cause code to fail in some (but not all) versions of python 3.

Instead we might want to consider dropping 3.1 and 3.2 (for example, but the exact versions can be up for discussion) since they seem less common. This will both limit the number of tests that we run (making development faster) and allow developers to take advantage of features added in later versions of python.

timj:
The specific version number is actually enforced in the `python` EUPS package and `newinstall.sh` to be >3.5. I think it's best if we don't mention a version in the style guide and let the system enforce the rules programmatically.

fred3m:
It might be useful to put this somewhere. For one, it helps developers to know which functions we can include in our python 3 code (the choice of 3.5 is great!). It can also be useful for debugging purposes.

fred3m:
An alternative solution would be to update the style guide to explicitly state the python 3 version is subject to change and can be found in EUPS. But somewhere we should let users know the requirement.

jdswinbank:
In RFC-14 we agreed that the versions of Python we support should be specified in the Developer Guide. That appears to have been dropped in the recent refactoring (DM-5456): I'm not sure if that was deliberate, or just an oversight.

coding/python_style_guide.rst
```diff
@@ -15,6 +15,29 @@ The :doc:`intro` provides the overarching Coding Standards policy applicable to
...
+=================
+
+.. _style-guide-py-version:
+
+All DM Python code MUST work with Python 3
+------------------------------------------
+
+All the Python code written by LSST Data Management must be runnable using Python 3.
+Python 2 will cease to be supported before LSST is operational (:pep:`373`).
+The current baseline version is Python 3.5 or above.
```
ktlim: A "current baseline" should never be a range.  Delete "or above", please.

coding/python_style_guide.rst
```diff
@@ -15,6 +15,29 @@ The :doc:`intro` provides the overarching Coding Standards policy applicable to
...
+
+DM Python library code with an external user base MUST support Python 2.7 and 3.x
+---------------------------------------------------------------------------------
+
+During construction we are expecting external users to be experimenting with some of the DM code and LSST DM library code is being used in their applications.
+This user community is currently transitioning from Python 2.7 to Python 3.x and we should ensure that code works on both Python versions until our dependencies drop support for Python 2.7.
```
ktlim: Comma before "and we".

coding/python_style_guide.rst
```diff
@@ -15,6 +15,29 @@ The :doc:`intro` provides the overarching Coding Standards policy applicable to
...
+In particular, the Science Pipelines code (commonly referred to as ``lsst_distrib``) must support 2.7 and 3.x.
+
+Standalone applications, code providing services, and internal programs and modules (such as Qserv, SQuaSH and ``dax``) do not have to support Python 2.
+If code is currently supporting both 2.7 and 3.x, dropping support for Python 2.7 requires an :ref:`RFC <decision-making-rfc>`.
+New code that has never supported Python 2.7 and which will not be externally usable library code or a dependency of a package that support 2.7, does not require an RFC to request that 2.7 is not supported.
```
ktlim: "support 2.7," -> "supports 2.7" (no comma)

coding/python_style_guide.rst
```diff
@@ -620,6 +643,40 @@ For more information see the online book `Supporting Python 3 <http://python3por
...
+Details on the process for migrating a 2.7 codebase to support both versions can be found in `SQR-014 <https://sqr-014.lsst.io>`_.
+
+.. _style-guide-py-future-absolute-import:
+.. _style-guide-py-future-division:
+.. _style-guide-py-print:
+
+``__future__`` MUST be used to import Python 3 behavior in all files
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Code that is to be used on Python 2.7 and 3.x should import ``division``, ``print_function`` and ``absolute_import`` from the :mod:`__future__` package.
```
ktlim: "MUST" vs. "should" is confusing.  Use "must" here but qualify with "where the related functionality is used"?

coding/python_style_guide.rst
```diff
@@ -620,6 +643,40 @@ For more information see the online book `Supporting Python 3 <http://python3por
...
+In addition, import local modules using relative imports (e.g. ``from . import foo`` or ``from .foo import bar``).
+This results in clearer code and avoids shadowing global modules with local modules.
+
+The :py:func:`print()` function is required in Python 3.
+In general, DM code should use logging instead of ``print`` statements.
+
+
+.. _style-guide-py-future-itervalues:
+
+``itervalues()`` and ``iteritems()`` CAN NOT be used
```
ktlim: "CANNOT" is generally preferred over "CAN NOT".

coding/python_style_guide.rst
timj: @fred3m @jdswinbank I do state python 3.5 here.

coding/python_style_guide.rst
```diff
@@ -15,6 +15,29 @@ The :doc:`intro` provides the overarching Coding Standards policy applicable to
...
+.. _style-guide-py-version:
+
+All DM Python code MUST work with Python 3
+------------------------------------------
+
+All the Python code written by LSST Data Management must be runnable using Python 3.
+Python 2 will cease to be supported before LSST is operational (:pep:`373`).
+The current baseline version is Python 3.5 or above.
+
+DM Python library code with an external user base MUST support Python 2.7 and 3.x
```
timj: @fred3m @jdswinbank I do state 2.7 here. So I think that's fine. The 3.x here is to stop duplicating the explicit version number stated above.

PR #62, opened 2016-09-06
Fix spelling: interable -> iterable

kfindeisen:
I saw the "interable" typo in one of the docstring examples. I assume I don't actually need to fill out an RFC for changes on this level...


jonathansick:
Thanks @kfindeisen!

PR #64, opened 2016-09-08
DM-7568: Fix link in C++ Style Guide to using devtoolset-3

jonathansick:
- The original page on confluence was deprecated by
  services/lsst-dev.rst
- Reformatted as a nice seealso block for clarity.


jonathansick:
See https://developer.lsst.io/v/DM-7568/coding/cpp_style_guide.html#style-guide-cpp-cpp-11-14

timj:
Can we link to the pipelines version of the instructions as well? Given that those are aimed more at developers own machines rather than using pre-configured lsst machines.

jonathansick:
Does the latest update work for you @timj ?

PR #65, opened 2016-09-09
Added setup script example for csh shells.

kfindeisen:
Having a bash and a csh example present should be sufficient, given
that `deploy` provides interactive instructions as well.


kfindeisen:
There seems to be some unwanted syntax highlighting in the second example, I'll have to figure out where it's coming from: https://developer.lsst.io/v/DM-7552/build-ci/lsstsw.html

PR #74, opened 2016-10-06
DM-7919: Add git config requirement for 'simple' pushes

jonathansick:
**Preview**

- https://developer.lsst.io/v/DM-7919/processes/workflow.html#git-setup
- https://developer.lsst.io/v/DM-7919/tools/git_setup.html

****

See https://community.lsst.org/t/the-dangers-of-git-on-lsst-dev/1216?u=jsick for background.

### DIFF
--- a/processes/workflow.rst
+++ b/processes/workflow.rst
@@ -26,21 +26,13 @@ Other related pages:
 Git & GitHub Setup
 ==================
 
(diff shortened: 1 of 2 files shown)


jonathansick:
@kfindeisen Thanks for taking a look!

PR #76, opened 2016-10-19
Specify shared-library variant of Boost.Test

jonathansick:
https://developer.lsst.io/v/DM-8038/coding/unit_test_policy.html

### DIFF
--- a/coding/unit_test_policy.rst
+++ b/coding/unit_test_policy.rst
@@ -79,7 +79,7 @@ Test suite execution should be managed by a testing framework, also known as a
 test harness, which monitors the execution status of individual test cases.
 
 C++: boost.test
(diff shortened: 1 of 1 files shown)


coding/unit_test_policy.rst
```diff
@@ -79,7 +79,7 @@ Test suite execution should be managed by a testing framework, also known as a
 test harness, which monitors the execution status of individual test cases.
 
 C++: boost.test
-    LSST DM developers should use the `single-header variant`_ of the `Boost Unit
+    LSST DM developers should use the `shared-header variant`_ of the `Boost Unit
```
kfindeisen: Should be shared library, not shared header (there's no such thing). Otherwise looks good.

jonathansick: Oh I see what you mean. I'll fix it and merge then.

PR #92, opened 2016-12-20
DM-7891: Add detailed C++ documentation guidelines

kfindeisen:
This is a rewrite of the C++ documentation guide to have similar format and content to the [Python documentation guide](https://developer.lsst.io/docs/py_docs.html). All changes discussed in [RFC-225](https://jira.lsstcorp.org/browse/RFC-225) have been included. However, any inconsistencies between the new text and the [Confluence documentation](https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards) have not been resolved; they will be addressed as part of [DM-7892](https://jira.lsstcorp.org/browse/DM-7892).

A compiled version of the new style guide is available at https://developer.lsst.io/v/DM-7891/docs/cpp_docs.html. The actual diff is quite messy, with `git diff` apparently using blank lines as reference points.

### DIFF
--- a/docs/cpp_docs.rst
+++ b/docs/cpp_docs.rst
@@ -190,6 +190,7 @@ For summaries of how these sections are composed in specific contexts, see:
 
 - :ref:`cpp-doxygen-package-definition`
 - :ref:`cpp-doxygen-class-structure`
+- :ref:`cpp-doxygen-enum-structure`
 - :ref:`cpp-doxygen-method-function-structure`
 - :ref:`cpp-doxygen-attribute-constants-structure`
 
@@ -587,8 +588,8 @@ before any documented code, and a Doxygen comment with a ``@}`` after any docume
 
 .. _cpp-doxygen-class-structure:
 
-Documenting Classes
-===================
+Documenting Classes and Type Aliases
+====================================
 
 Class documentation blocks are placed immediately before the class declaration, and serve to document the class as a whole rather than individual methods.
 
@@ -619,6 +620,67 @@ For example:
        public:
            ...
    }
+
+Type alias declarations and typedefs should also be documented, although just a short summary is usually sufficient.
+Doxygen will automatically provide links to the types being renamed, if their documentation is available.
+
+.. _cpp-doxygen-enum-structure:
+
+Documenting Enumerated Types
+============================
+
+An enumerated type is a type, and should be documented similarly to a class:
+
+1. :ref:`cpp-doxygen-short-summary`
+2. :ref:`cpp-doxygen-deprecation` (if applicable)
+3. :ref:`cpp-doxygen-extended-summary` (recommended)
+4. :ref:`cpp-doxygen-see-also` (optional)
+5. :ref:`cpp-doxygen-notes` (optional)
+6. :ref:`cpp-doxygen-references` (optional)
+
+In addition, each value of the type should be documented. A short description is almost always sufficient.
+
+For example:
+
+.. code-block:: cpp
+
+   /**
+    * Supported coordinate systems for flux-conserving transformations.
+    *
+    * These values are used in arguments to various functions in this package.
+    * Unless otherwise stated, these functions do not validate whether the data
+    * set makes sense in the "from" coordinates.
+    */
+   enum class CoordType
+   {
+       /// Untransformed detector coordinates.
+       PIXEL,
+       /// Idealized detector coordinates after applying a distortion correction.
+       WARP_PIXEL,
+       /// Equatorial J2000.0 coordinates.
+       SKY_WCS
+   };
+
+.. _cpp-doxygen-enum-inline:
+
+Annotating Enum Values with Inline Comments (optional)
+------------------------------------------------------
+
+If the value descriptions are very short, you may choose to annotate values with inline comments after each constant, one per line.
+These comments are prefixed with ``///<``.
+
+For example:
+
+.. code-block:: cpp
+
+   enum class CoordType
+   {
+       PIXEL,    ///< Untransformed detector coordinates
+       WARP_PIXEL,    ///< Distortion-corrected detector coordinates
+       SKY_WCS    ///< Equatorial J2000.0 coordinates
+   };
+
+If the constant descriptions are too long to fit in a single line of source, ordinary documentation blocks before each constant must be used.
 
 .. _cpp-doxygen-method-function-structure:
 
--- a/docs/cpp_docs.rst
+++ b/docs/cpp_docs.rst
@@ -549,38 +549,57 @@ Examples should be prefixed by ``@code`` and suffixed by ``@endcode``:
 
 .. _cpp-doxygen-package-definition:
 
-Package Documentation / Definition
-==================================
+Documenting/Defining Packages
+=============================
 
 Each LSST package corresponds to a group in Doxygen.
 We declare this package in the root header file for a package, usually named ``package.h``.
 
-In this header file, below the preamble, provide a Doxygen comment block that declares the package with the fields:
+In this header file, below the boilerplate but above any ``#include`` statements or other code, provide a Doxygen comment block that declares the package with the fields:
 
-1. ``@defgroup`` followed by the ``PackageName`` and  ``PackageTitle``
-
-2. ``@brief`` to provide a one-line description of the package.
+1. ``@defgroup`` followed by machine-readable and human-readable names for the package
+2. :ref:`cpp-doxygen-short-summary` (must be prefixed by ``@brief`` because the summary is not the first line.)
+3. :ref:`cpp-doxygen-deprecation` (if applicable)
+4. :ref:`cpp-doxygen-extended-summary` (recommended)
+5. :ref:`cpp-doxygen-see-also` (optional)
 
 For example:
 
 .. code-block:: cpp
 
    /**
-    * @defgroup PackageName PackageTitle
+    * @defgroup afw LSST Framework
     *
-    * @brief Provide some stuff to do stuff
+    * @brief LSST data management: astronomical framework, including images and tables
     */
+
+Header files that belong to the package should have a Doxygen comment like
+
+.. code-block:: cpp
+
+   /**
+    * @addtogroup group
+    *
+    * @{
+    */
+
+before any documented code, and a Doxygen comment with a ``@}`` after any documented code. Everything between the ``@{`` and the ``@}`` will be listed on the module page.
 
 .. _cpp-doxygen-class-structure:
 
-Class Definitions
-=================
+Documenting Classes
+===================
 
-Where a class is *defined* (usually in a header file), provide a Doxygen block preceeding the class that includes
+Class documentation blocks are placed immediately before the class declaration, and serve to document the class as a whole rather than individual methods.
 
-1. A one-line description of the class.
-
-2. A paragraph (or more) describing the class. Markdown can be used to provide nuanced typography.
+1. :ref:`cpp-doxygen-short-summary`
+2. :ref:`cpp-doxygen-deprecation` (if applicable)
+3. :ref:`cpp-doxygen-extended-summary` (recommended)
+4. :ref:`cpp-doxygen-tparameters` (if applicable)
+5. :ref:`cpp-doxygen-see-also` (optional)
+6. :ref:`cpp-doxygen-notes` (optional)
+7. :ref:`cpp-doxygen-references` (optional)
+8. :ref:`cpp-doxygen-examples` (optional)
 
 For example:
 
@@ -595,59 +614,120 @@ For example:
     * list "utils.dlist", and the code to destroy a list "utils.dlist.del"
     *
     */
-   class TraceImpl {
+   class TraceImpl
+   {
        public:
+           ...
    }
 
 .. _cpp-doxygen-method-function-structure:
 
-Function/Method Definitions
-===========================
+Documenting Methods and Functions
+=================================
 
-Where a function or class method is *defined*, provide a Doxygen block preceeding that class that includes
+All public or protected methods and all functions must be preceded by a documentation block.
+Method or function documentation blocks contain the following sections:
 
-1. A one-line description of the function/method
+1. :ref:`cpp-doxygen-short-summary`
+2. :ref:`cpp-doxygen-deprecation` (if applicable)
+3. :ref:`cpp-doxygen-extended-summary` (recommended)
+4. :ref:`cpp-doxygen-tparameters` (if applicable)
+5. :ref:`cpp-doxygen-parameters` (if applicable)
+6. :ref:`cpp-doxygen-returns` (if applicable)
+7. :ref:`cpp-doxygen-throws` (if applicable)
+8. :ref:`cpp-doxygen-exceptsafe` (optional)
+9. :ref:`cpp-doxygen-related` (if applicable; for functions only)
+10. :ref:`cpp-doxygen-see-also` (optional)
+11. :ref:`cpp-doxygen-notes` (optional)
+12. :ref:`cpp-doxygen-references` (optional)
+13. :ref:`cpp-doxygen-examples` (optional)
 
-2. Optionally, a paragraph or more with detailed descriptions of the function/method. Markdown can be used here.
-
-3. ``@param`` statements describing each function/method argument. Optionally, inline comments can be used (see below).
-
-An example of a Doxygen comment for a function:
+An example:
 
 .. code-block:: cpp
 
-   /** Set a component's verbosity.
-   *
-   * If no verbosity is specified, inherit from parent
-   *
-   * @param name component of interest
-   * @param verbosity desired trace verbosity
-   */
-   void TraceImpl::setVerbosity(const std::string &name, const int verbosity) {
-   }
+   /**
+    * Read an image from disk.
+    *
+    * @param fileName the file to read. Must be either absolute or relative to
+    *     the program working directory.
+    *
+    * @return the image stored in `fileName`. If the image on disk does not
+    *     have `double` pixels, they will be cast to `double`.
+    *
+    * @throws IoError Thrown if `fileName` does not exist or is not readable.
+    *
+    * @exceptsafe Strong exception guarantee.
+    */
+   lsst::afw::image::Image<double> loadImage(std::string const & fileName);
 
-Overloaded Function/Methods Definitions
-=======================================
+.. _cpp-doxygen-method-function-overloads:
 
-'`@overload`` may be used when two methods/functions are effectively the same but have different parameters list for reasons of convenience.
+Overloaded Function/Method Definitions
+--------------------------------------
+
+``@overload`` may be used when two methods/functions are effectively the same but have different parameter lists for reasons of convenience.
+Use this tag **only** when the specification of the abbreviated overload can be easily inferred from the fully documented one.
+
+The text generated by the ``@overload`` tag tells readers to see the method "above".
+Because Doxygen sorts the detailed documentation of namespace and class members, you should check the generated documentation to make sure the fully documented overload appears before any that use the ``@overload`` tag.
 
 For example:
 
 .. code-block:: cpp
 
    /**
-    * seconds from midnight
+    * Sum numbers in a vector.
+    *
+    * @param values Container whose values are summed.
+    * @return sum of `values`, or 0.0 if `values` is empty.
+    *
+    * @exceptsafe This function does not throw exceptions.
     */
-   long GetTime(void){
-       return secondFromMidnight(CURRENT);
-   }
+   double add(std::vector<double> const & values);
+
    /**
-    # @overload void GetTime(int &hours, int &minutes, int &seconds)
+    * Sum numbers in an array.
+    *
+    * @overload double add(std::vector<double> const &)
     */
-   void GetTime(int &hours, ///< set: current hour
-                int &minutes, ///< set: current minutes
-                int &seconds) { ///< set: current seconds
-    hours = _hours;
-    minutes = _minutes;
-    seconds = _seconds;
-   }
+   double add(double[] const values, size_t nValues);
+
+.. _cpp-doxygen-attribute-constants-structure:
+
+Documenting Constants, Variables, and Data Members
+==================================================
+
+All non-private constants, variables, or data members must be preceded by a documentation block.
+At minimum, constants/variables/data members should have a summary line, but can also have a more complete structure:
+
+1. :ref:`cpp-doxygen-short-summary`
+2. :ref:`cpp-doxygen-deprecation` (if applicable)
+3. :ref:`cpp-doxygen-extended-summary` (optional)
+4. :ref:`cpp-doxygen-initializer` (optional; for constants only)
+5. :ref:`cpp-doxygen-notes` (optional)
+6. :ref:`cpp-doxygen-references` (optional)
+7. :ref:`cpp-doxygen-examples` (optional)
+
+For example:
+
+.. code-block:: cpp
+
+   /// Flag set if background subtraction should not be done.
+   const int NO_BACKGROUND = 1 << 3;
+
+.. _cpp-doxygen-attribute-constants-inline:
+
+Annotating Constants and Variables with Inline Comments (optional)
+------------------------------------------------------------------
+
+If the constant, variable, or data member descriptions are very short, you may choose to annotate them with inline comments after each value, one per line.
+These comments are prefixed with ``///<``.
+
+For example:
+
+.. code-block:: cpp
+
+   const int NO_BACKGROUND = 1 << 3;        ///< Skip background subtraction
+
+If the descriptions are too long to fit in a single line of source, ordinary documentation blocks before each value must be used.
--- a/docs/cpp_docs.rst
+++ b/docs/cpp_docs.rst
@@ -7,16 +7,35 @@ Documenting C++ Code
    This is a preview documentation format specification.
    Software documentation should currently be written in the format described at https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards
 
-The LSST Stack uses `Doxygen <http://www.stack.nl/~dimitri/doxygen/>`_ to build C++ API reference documentation by extracting source code comments.
-This page covers the most important concepts in writing Doxygen-friendly C++ comments for the LSST Stack, though `Doxygen's manual <http://www.stack.nl/~dimitri/doxygen/manual.html>`_ is the most comprehensive reference.
 
-Note that LSST uses `Markdown-formatted Doxygen comment blocks <http://www.doxygen.nl/manual/markdown.html>`_.
+We document C++ code in two ways:
+
+1. By writing *documentation blocks* for all public or protected C++ components (namespaces, types, methods, functions, and constants).
+
+   The LSST Stack uses `Doxygen <http://www.doxygen.org/>`_ to build C++ API reference documentation from comment blocks. This documentation is exposed to users in a variety of contexts, from developers reading the code to readers of the `Stack Documentation <https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/>`_.
+
+   Doxygen comment blocks are the public specification of our C++ API.
+
+2. By commenting our code internally with C++ comments (``//`` or ``/* .. */``).
+
+   These comments are meant to be read only by developers reading and editing the source code.
+
+This page focuses on public code documentation using Doxygen, while internal comments are discussed in our :doc:`../coding/cpp_style_guide`.
+
+Treat the guidelines on this page as an extension of the :doc:`../coding/cpp_style_guide`.
+
+.. note::
+
+   Changes to this document must be approved by the System Architect (`RFC-24 <https://jira.lsstcorp.org/browse/RFC-24>`_).
+   To request changes to these standards, please file an :ref:`RFC <decision-making-rfc>`.
 
 This document is ported from original documentation in Confluence; https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards.
-**This guide will evolve as a mechanism to document C++ APIs wrapped in Python in developed.**
+**This guide will evolve as a mechanism to document C++ APIs wrapped in Python is developed.**
 
-Boilerplate for Header (.h) and Source (.cpp) Files
-===================================================
+.. _cpp-file-boilerplate:
+
+Boilerplate for Header (.h) and Source (.cc) Files
+==================================================
 
 The beginning of both header and source code files should include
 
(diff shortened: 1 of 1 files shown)


docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+    */
+
+Do not use a space between the parameters.
+
+.. _cpp-doxygen-parameters:
+
+Function/Method Parameters
+--------------------------
+
+A series of ``@param`` tags, usually one for each parameter. Each tag should have a description following the parameter name. You do *not* usually need to document default arguments; Doxygen will provide the default automatically. If the description extends over multiple lines, each line after the first must be indented.
```
TallJimbo: Should specify how it should be indented.  Example below suggests additional should line up with the parameter name.  Note that we more frequently line up with the start of the description, which is cleaner but can lead to many more line breaks when parameter names are long.  I'm ambivalent about which one we go with.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+Do not use a space between the parameters.
+
+.. _cpp-doxygen-parameters-inline:
+
+Annotating Parameters with Inline Comments (optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the parameter descriptions are very short, you may choose to annotate parameters with inline comments after each parameter, one per line.
+These comments are prefixed with ``///<``.
```
TallJimbo: FWIW, I'd personally prefer to just disallow this form, but I think there may be some strong advocates elsewhere one the team.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+Doxygen will render one or more ``@throws`` tags as a table of exceptions and descriptions, so do not treat ``@throws`` as the first word of the description.
+
+For consistency with Python documentation, always use ``@throws`` and not the synonymous ``@throw`` or ``@exception``.
+
+.. _cpp-doxygen-exceptsafe:
+
+Exception Safety
+----------------
+
+Whether or not there are any ``@throws`` tags, a function or method must have an ``@exceptsafe`` tag. (This applies even if the function or method is declared ``noexcept``, because the ``noexcept`` modifier is easy to miss in the generated documentation.) The description following the tag should describe the level of exception safety provided by the function or method.
```
TallJimbo: Our exception safety documentation is currently so terrible that I'm not sure it's worthwhile to mandate this in the most common case of essentially no exception safety, since most functions will call existing lower-level functions that don't document this at all.

I do think it'd be in our best interest to clean that up, but I don't think adding `@exceptsafe May throw arbitrary exceptions, leaving the program in an indeterminate state.` to virtually every function is worthwhile in the meantime.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+.. _cpp-doxygen-notes:
+
+Notes
+-----
+
+*Notes* is an optional section that provides additional information about the code, possibly including a discussion of the algorithm or known limitations of the code. The notes should be prefixed by a ``@note`` or ``@warning`` command.
+
+For internal consistency, always use ``@note`` and not the synonymous ``@remark`` or ``@remarks``.
+
+This section may include mathematical equations, written in `LaTeX <http://www.latex-project.org/>`_ format:
```
TallJimbo: This implies that the extended documentation section should not contain Latex math, which I think we don't want to discourage; many of our functions and classes are most easily described by referring to the mathematics of the algorithm they represent.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+Math can also be used inline:
+
+.. code-block:: cpp
+
+   /**
+    * The value of @f$\omega@f$ is larger than 5.
+    */
+
+Note that LaTeX is not particularly easy to read, so use equations sparingly. Doxygen recovers poorly from typos in formulas; you may need to manually delete ``docs/html/formula.repository`` if it contains a bad formula.
```
TallJimbo: I slightly disagree with this.  Latex markup is indeed hard to read, but it can make the rendered documentation vastly easier to read, especially for the hardest-to-explain code objects.  Perhaps we could reword this to discourage gratuitous use of Latex math (e.g. to get script Greek letters) while encouraging it for highly mathematical code objects?

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
 
-1. ``@defgroup`` followed by the ``PackageName`` and  ``PackageTitle``
+1. ``@defgroup`` followed by machine-readable and human-readable names for the package
+2. :ref:`cpp-doxygen-short-summary` (must be prefixed by ``@brief`` because the summary is not the first line.)
+3. :ref:`cpp-doxygen-deprecation` (if applicable)
+4. :ref:`cpp-doxygen-extended-summary` (recommended)
+5. :ref:`cpp-doxygen-see-also` (optional)
 
-2. ``@brief`` to provide a one-line description of the package.
+The extended summary should include the `preferred abbreviation <https://developer.lsst.io/coding/cpp_style_guide.html#names-representing-namespaces-must-be-camelcase-with-leading-lowercase-letter-and-based-on-component-or-directory-name>`_ for the package's namespace, if one exists.
```
TallJimbo: Maybe a question for @jonathansick: how does this package summary relate to the Python summary?  It seems like they'd contain essentially the same content, and should perhaps fill in one from the other?

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+File Description Comment for Header Files
+=========================================
+
+``.h`` files should not contain documentation on the file itself (i.e., a documentation block containing a ``@file`` command). Document the class or package instead, where possible.
+
+.. _cpp-doxygen-package-definition:
+
+Documenting/Defining Packages
+==============================
```
TallJimbo: I realize this was in place before this rewrite, but I wonder if we should stop using groups to represent packages in Doxygen.  Our namespaces are already 1-1 with our packages, so we could put this summary information in the namespace docs, and using groups results in a fair amount of boilerplace for every header file.

If we do decide to keep using groups for packages, we should perhaps move the recommendations for what to add to each header file to the `Boilerplate for Header and Source Files` section.

docs/cpp_docs.rst
```diff
@@ -128,77 +608,131 @@ For example:
...
-   }
+    *
+    * If no verbosity is specified, it will be inherited from its parent.
+    *
+    * @param name component of interest
+    * @param verbosity desired trace verbosity
+    *
+    * @exceptsafe This method shall leave the component unchanged in the event of an exception.
+    */
+   void setVerbosity(std::string const & name, int verbosity);
```
TallJimbo: To make this more help as a quick reference, it might be a good idea to add (and document) a return value and some possible exceptions that could be thrown here.

docs/cpp_docs.rst
```diff
@@ -7,15 +7,33 @@ Documenting C++ Code
    This is a preview documentation format specification.
    Software documentation should currently be written in the format described at https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards
```
kfindeisen: I'm hoping to do that in DM-7892, but AFAIK there is still be useful material on that page that needs to be ported over.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+    * @param values Container whose values are summed.
+    * @return sum of `values`, or 0.0 if `values` is empty.
+    */
+
+.. _cpp-doxygen-refs:
+
+Cross-References MUST be prefixed by `@ref`, and SHOULD be used sparingly
+--------------------------------------------------------------------------
+
+References to other classes, functions, and methods will only be linked if they are prefixed by the ``@ref`` tag. Excessive use of links can make the compiled documentation difficult to read; it is recommended that only the first mention of a particular component be a ``@ref``.
```
kfindeisen: The current Doxygen config has autolinking disabled: https://github.com/lsst/base/blob/master/doc/base.inc#L289.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+=====================================
+
+This page covers the most important concepts in writing Doxygen-friendly C++ comments for the LSST Stack, though `Doxygen's manual <http://www.doxygen.org/manual/>`_ is the most comprehensive reference.
+
+.. _cpp-doxygen-javadoc:
+
+Documentation MUST be delimited in Javadoc style
+-------------------------------------------------
+
+Documentation blocks **must** begin with ``/**`` and end in ``*/``. The only exception is for :ref:`inline parameter documentation <cpp-doxygen-parameters-inline>`, which may begin with ``///<``.
```
TallJimbo: I have no ab initio preference between `///` and `/** ... */`, but I think the former is more common in the codebase.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+----------------
+
+A few sentences giving an extended description.
+This section should be used to clarify *functionality*, not to discuss implementation detail or background theory, which should rather be explored in the :ref:`cpp-doxygen-notes` section below.
+You may refer to the parameters and the function name, but parameter descriptions still belong in the :ref:`cpp-doxygen-parameters` section unless they are very lengthy.
+
+.. _cpp-doxygen-tparameters:
+
+Template Parameters
+-------------------
```
kfindeisen: Oops, good catch. Yes, treating these cases in a manner similar to `@overload` seems reasonable.

docs/cpp_docs.rst
kfindeisen: Oops, I thought I'd stuck to 4-space indentation in all my examples.

Since this isn't Python, it didn't seem as important to stick to an exact indentation style, as long as it's human-readable. I personally also use start-of-description alignment, but I didn't want to require or imply it because it is more work when writing/editing docs.

docs/cpp_docs.rst
kfindeisen: Again, the wording was what it was for consistency with the Python documentation. In this case, though, I think it's also an intrinsically good idea -- unless a particular class is highly mathematical in nature, recommending the use of math in the main description may encourage documentation-writers to focus on implementation details rather than an appropriate level of abstraction.

docs/cpp_docs.rst
kfindeisen: I was wondering why nobody objected to this bit at the RFC stage; I expected it to be the most controversial part of the proposal (granted, it was only one line in a page-long document).

I agree that much of the existing code is almost certainly not exception-safe. My hope was that requiring this in the documentation would encourage developers to pay attention to exception safety when writing new code, whereas making it optional would cause the issue to get quietly ignored. I took it for granted that nobody would be going through and systematically rewriting old documentation the way you describe.

docs/cpp_docs.rst
kfindeisen: I think that's basically what "use equations sparingly" means. Is there some specific wording that would make it clearer?

docs/cpp_docs.rst
kfindeisen: I agree 100%. However, I think the group vs. namespace issue is big enough to merit its own RFC, and I didn't want it bogging down this one. The change here was a quick fix to work around the disappearance of the file-level comments, which previously contained an `@ingroup` statement.

docs/cpp_docs.rst
kfindeisen: Perhaps modifying an existing example based on real code I'm not familiar with was not the best way to approach this.

Though in this case, I actually wanted to have an example of an `@exceptsafe` tag without a corresponding `@throws` tag -- I've run into the situation where I don't know what exceptions will be thrown (because a function used in the implementation doesn't document them), but I can still use statement ordering and RAII to ensure exception safety.

docs/cpp_docs.rst
jonathansick: I'm wondering how much documentation needs to be written in doxygen in this context. I think we can skip the Extended summary for packages since most documentation at this level will be written directly in Sphinx rST — I'm writing up a description of the Science Pipelines documentation architecture in DMTN-030 as we speak after a meeting we had a week and a half ago.

jonathansick:
This all looks fantastic. Thanks for doing this @kfindeisen. I think we can go ahead with this (given other's comments of course). My main concerns that can probably get worked out in separate tickets are:

- What changes are needed to the Doxygen config to make this work? It'd be nice to a rendered example to know it'll all work.
- There may relationship between this work and how Pybind11-wrapped Python APIs are documented. The consistency established in this ticket will certainly help, but there may be unknown adjustments needed to our API documentation specs in the future.
- Similarly, I need to work out the impact of this spec on Doxygen bridged with [Breathe](https://breathe.readthedocs.io/en/latest/) into Sphinx. My guess is that this PR is entirely beneficial, but again some adjustments may be necessary.

docs/cpp_docs.rst
jonathansick: Yeah, I think what @kfindeisen says is the spirit of it. That recommendation came directly from Numpydoc itself, but we can adjust the language in both style guides to be more explicit about good and bad form.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+For functions and methods, the summary should be written in the imperative voice (i.e., as a command that the API consumer is giving). Getters and other methods that are more naturally described as values rather than actions may ignore this rule.
+
+.. _cpp-doxygen-deprecation:
+
+Deprecation Warning
+-------------------
+
+A ``@deprecated`` tag (where applicable) to warn users that the component is deprecated. The text that follows the tag should include:
+
+1. In what stack version the object was deprecated, and when it will be removed.
```
TallJimbo: Leaving this as-is and updating both it and the Python guide following RFC-213 is fine with me, though I have a slight preference for updating both to not recommend something we can't really do for now.

docs/cpp_docs.rst
TallJimbo: At least for me, having it there even if it's optional is still a helpful reminder to think about exception safety, though I take your point that it's not nearly as helpful in that regard if it's optional.  I'm mostly worry about adding a ton of documentation boilerplate to simple functions whose exception safety is determined entirely by calls to other less-documented code.

Could we define (here in the docs) a few labels to allow the `@exceptsafe` tags to be extremely brief (e.g. ["noexcept", "strong", "basic", "none"](https://en.wikipedia.org/wiki/Exception_safety))?

docs/cpp_docs.rst
TallJimbo: Making it more explicit sounds like a good idea.  How about:
 - Do not hesitate to use Latex math to describe the behavior of a code object, but be sure to put math that pertains only to the implementation in `@note` sections rather than the extended description.
 - Do not use inline Latex math just to add Greek or other special symbols, unless the documentation block involves longer display-style equations using the same symbols.  (Use unicode instead?)

docs/cpp_docs.rst
TallJimbo: Deferring this to another RFC is fine with me; I don't expect that to be a controversial one.  @kfindeisen, are you planning to put that RFC together?

kfindeisen:
@jonathansick This writeup is based almost entirely on existing Doxygen settings, except for [DM-7893](https://jira.lsstcorp.org/browse/DM-7893). The [SpherePoint class](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/classlsst_1_1afw_1_1geom_1_1_sphere_point.html) was documented following an early version of these rules; I think the only features that are missing are equations and figures.

docs/cpp_docs.rst
```diff
@@ -7,15 +7,33 @@ Documenting C++ Code
...
+   Changes to this document must be approved by the System Architect (`RFC-24 <https://jira.lsstcorp.org/browse/RFC-24>`_).
+   To request changes to these standards, please file an :ref:`RFC <decision-making-rfc>`.
 
 This document is ported from original documentation in Confluence; https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards.
 **This guide will evolve as a mechanism to document C++ APIs wrapped in Python in developed.**
 
-Boilerplate for Header (.h) and Source (.cpp) Files
+.. _cpp-file-boilerplate:
+
+Boilerplate for Header (.h) and Source (.cc) Files
```
jonathansick: Make sure the header's underline matches the length of the text.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+       * You should have received a copy of the LSST License Statement and
+       * the GNU General Public License along with this program. If not,
+       * see <http://www.lsstcorp.org/LegalNotices/>.
+       */
 
-File Description Comment for Header Files
-=========================================
+.. _cpp-doxygen-basics:
+
+Basic Format of Documentation Blocks
```
jonathansick: Also check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+=====================================
+
+This page covers the most important concepts in writing Doxygen-friendly C++ comments for the LSST Stack, though `Doxygen's manual <http://www.doxygen.org/manual/>`_ is the most comprehensive reference.
+
+.. _cpp-doxygen-javadoc:
+
+Documentation MUST be delimited in Javadoc style
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
 
-After the boilerplate, any C++ header file should have a Doxygen-style comment block (that begins with ``/**``) that includes the following fields:
+For consistency, do not begin documentation blocks with ``/*!``.
 
-1. ``@file`` giving the file's name
+.. _cpp-doxygen-tag:
 
-2. ``@brief`` to provide a synopsis of the file for Doxygen's index. This should be a line, at most.
+Documentation MUST use Javadoc-style tags
+------------------------------------------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
 
-3. ``@ingroup`` to specify the name of the LSST Stack package containing this file (see :ref:`below <doc-cpp-package-definition>`).
+Documentation blocks must use tags such as ``@see`` or ``@param`` in place of ``\see`` or ``\param``.
+This is both for internal consistency and to avoid conflicts with other tools that give special treatment to ``\word``.
 
-4. ``@author`` to provide the name of the file's primary author.
+.. _cpp-doxygen-form:
+
+Documentation delimiters SHOULD be on their own lines
+------------------------------------------------------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+    * mathematicians have heard of.
+    *
+    * @param values Container whose values are summed.
+    * @return sum of `values`, or 0.0 if `values` is empty.
+    */
+
+.. _cpp-doxygen-refs:
+
+Cross-References MUST be prefixed by `@ref`, and SHOULD be used sparingly
+--------------------------------------------------------------------------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
jonathansick: What do you mean by "make the compiled documentation difficult to read?" As in, there are too many highlighted links? I wonder if we can fix whatever issue you have once the doxygen content is passed through Sphinx.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+References to other classes, functions, and methods will only be linked if they are prefixed by the ``@ref`` tag. Excessive use of links can make the compiled documentation difficult to read; it is recommended that only the first mention of a particular component be a ``@ref``.
+
+.. _cpp-doxygen-styling:
+
+Documentation SHOULD use Markdown for formatting
+-------------------------------------------------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+LSST uses `Markdown-formatted Doxygen comment blocks <http://www.doxygen.org/manual/markdown.html>`_. If a particular format cannot be expressed using Markdown, you MAY use `Doxygen's built-in formatting <http://www.doxygen.org/manual/commands.html>`_ or, if necessary, `HTML markup <http://www.doxygen.org/manual/htmlcmds.html>`_.
+
+.. _cpp-doxygen-headeronly:
+
+Documentation MUST appear where a component is first declared
+--------------------------------------------------------------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+In general, this means documentation blocks will appear in header (``.h``) files rather than source (``.cc``) files. This goes against Doxygen convention, but avoids certain false alarms when Doxygen parses C++11 code.
+
+.. _cpp-doxygen-indentation:
+
+Documentation MUST appear before the declaration it describes, and with the same indentation
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+
+.. _cpp-doxygen-deprecation:
+
+Deprecation Warning
+-------------------
+
+A ``@deprecated`` tag (where applicable) to warn users that the component is deprecated. The text that follows the tag should include:
+
+1. In what stack version the object was deprecated, and when it will be removed.
+2. Reason for deprecation if this is useful information (e.g., object is superseded, duplicates functionality found elsewhere, etc.).
```
jonathansick: With the `e.g.,` you can leave out the `etc.`

(I need fix this in https://developer.lsst.io/docs/py_docs.html#deprecation-warning 😬!)

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+-------
+
+A ``@returns`` tag, followed by a description similar to the one for :ref:`cpp-doxygen-parameters`. If the returned value is a map, ensure that the key-value pairs are documented in the description.
+
+For consistency with Python documentation, always use ``@returns`` and not the synonymous ``@return``.
+
+.. _cpp-doxygen-throws:
+
+Throws
+-------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+    */
+
+where ``filename`` is a path relative to the project root directory.
+
+.. _cpp-doxygen-references:
+
+References
+----------
+
+References can be included either in the :ref:`'Notes' <cpp-doxygen-notes>` section or in a separate list below them. A reference consists of the ``@cite`` tag, followed by a BibTeX label. Bibfiles must be listed in the ``CITE_BIB_FILES`` configuration tag in ``doc/doxygen.conf.in``.
```
jonathansick: Hmm. I wonder if we should have a stub bibtex file in the `doc/` directories of every package to help make this consistent. Out of scope for this ticket, but I'll have to consider this.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+    */
+
+File Description Comment for Header Files
+=========================================
+
+``.h`` files should not contain documentation on the file itself (i.e., a documentation block containing a ``@file`` command). Document the class or package instead, where possible.
+
+.. _cpp-doxygen-package-definition:
+
+Documenting/Defining Packages
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
jonathansick: Right, I'm not sure we doxygen documentation for packages. Fine if it's useful for developers working in-situ in the codebase, but I think this level of documentation will happen in https://dmtn-030.lsst.io/#module-topic-type, beyond the scope of doxygen.

docs/cpp_docs.rst
```diff
@@ -26,94 +44,556 @@ The beginning of both header and source code files should include
...
+    *
+    * @{
+    */
+
+before any documented code, and a Doxygen comment with a ``@}`` after any documented code. Everything between the ``@{`` and the ``@}`` will be listed on the module page.
+
+.. _cpp-doxygen-class-structure:
+
+Documenting Classes and Type Aliases
+=====================================
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -128,77 +608,131 @@ For example:
...
-Where a function or class method is *defined*, provide a Doxygen block preceeding that class that includes
+Type alias declarations and typedefs should also be documented, although just a short summary is usually sufficient.
+Doxygen will automatically provide links to the types being renamed, if their documentation is available.
 
-1. A one-line description of the function/method
+.. _cpp-doxygen-enum-structure:
 
-2. Optionally, a paragraph or more with detailed descriptions of the function/method. Markdown can be used here.
+Documenting Enumerated Types
+=============================
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -128,77 +608,131 @@ For example:
...
+       /** Idealized detector coordinates after applying a distortion correction. */
+       WARP_PIXEL,
+       /** Equatorial J2000.0 coordinates. */
+       SKY_WCS
+   };
+
+.. _cpp-doxygen-method-function-structure:
+
+Documenting Methods and Functions
+==================================
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -128,77 +608,131 @@ For example:
...
+    *
+    * @exceptsafe This method shall leave the component unchanged in the event of an exception.
+    */
+   void setVerbosity(std::string const & name, int verbosity);
+
+.. _cpp-doxygen-method-function-overloads:
 
-If the argument descriptions are too long to fit in a single line of source, the ``@param`` documenation method should be used.
+Overloaded Function/Method Definitions
+---------------------------------------
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
```diff
@@ -128,77 +608,131 @@ For example:
...
-    hours = _hours;
-    minutes = _minutes;
-    seconds = _seconds;
-   }
+   double add(double[] const values, size_t nValues);
+
+.. _cpp-doxygen-attribute-constants-structure:
+
+Documenting Constants, Variables, and Data Members
+===================================================
```
jonathansick: Check the underline length here.

docs/cpp_docs.rst
jonathansick: At this stage I'd be tempted to just drop the warning :) Do we know what relevant material is left on that confluence page?

docs/cpp_docs.rst
kfindeisen: Yes, too many highlighted links. I had assumed this was the intent of disabling autolinking in the first place.

In any case, this is more an HTML/presentation issue than a Doxygen vs. Sphinx issue; I grabbed this rule from the [Javadoc style guidelines](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide) even though we'll obviously never use Javadoc.

docs/cpp_docs.rst
r-owen: I agree

kfindeisen: Could we get away with saying "For historical reasons, this format is permitted but not encouraged, and should not be used in new code."? There's some wording like that in the main style guide...

@r-owen I just added an [example of a documented constant](https://developer.lsst.io/v/DM-7891/docs/cpp_docs.html#documenting-constants-variables-and-data-members) using the `/* ... */` style rather than `///` or `///<`. I admit the latter would save some space...

TallJimbo: That works for me.

docs/cpp_docs.rst
jonathansick: Okay, I see. I think this runs counter to the direction I want to go with our docs (that all API entities are always consistently hyperlinked), and I acknowledge that it'll be a design consideration to make this clean.

I'm fine if we leave this rule in for now, but I may push for amendment when our C++ API docs are run through Sphinx and I'm developing the visual presentation of C++ docs.

docs/cpp_docs.rst
r-owen: I would personally rather use `///<` for documenting constants and such. Space is usually at a premium and I personally dislike using `/*...*/` (or variants such as `/**...*/` for comments that are intrinsically one-liners. It feels needlessly unsafe.

But that's just my opinion.

kfindeisen: Ok, I'm clearly getting outvoted on this. So is the consensus style:
- General comments:
    - multiline comments must be `/** */`
    - single-line comments must be `///`
- Function parameters:
    - Postfix comments (`///<`) allowed but discouraged in favor of `@param`
- Enum values and constants/variables:
    - Prefix comments (`///`) and postfix comments (`///<`) both allowed

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+   {
+       ...
+   }
+
+When two or more consecutive template parameters have *exactly* the same description, they can be combined:
+
+.. code-block:: cpp
+
+   /**
+    * @param T,U the types of the pair components
```
ktlim: This should be `@tparam`, no?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+
+   /**
+    * Compute mean and standard deviation for a collection of data.
+    *
+    * @param[out] mean the mean of `data`, or `NaN` if `data` is empty
+    * @param[out] stdDev the unbiased (sample) standard deviation, or `NaN`
+    *     if `data` contains fewer than 2 elements
+    * @param[in] data the data to analyze
+    */
+   std::ostream operator<<(double & mean, double & stdDev, std::vector<double> const & data);
```
ktlim: Is `operator<<` really a good example here?  Seems kind of strange.

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+Doxygen will render one or more ``@throws`` tags as a table of exceptions and descriptions, so do not treat ``@throws`` as the first word of the description.
+
+For consistency with Python documentation, always use ``@throws`` and not the synonymous ``@throw`` or ``@exception``.
+
+.. _cpp-doxygen-exceptsafe:
+
+Exception Safety
+----------------
+
+Whether or not there are any ``@throws`` tags, a function or method should have an ``@exceptsafe`` tag.
```
ktlim: I'm worried that this is too much to expect people to document, particularly if it involves researching all functions called by this function.  I also see that this is "recommended" for functions, which, while technically optional, seems too strong to me, particularly in combination with the "should" here.  I would allow omitting this, which by default would mean "no guarantees -- might be any of the cases below".

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+
+.. _cpp-doxygen-tparameters-specializations:
+
+Template Specializations
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When a partial template specialization reuses parameters from the full template, there is no need to redocument each parameter. If you are omitting the parameters, the documentation must include a cross-reference to the full template, possibly as part of the :ref:`cpp-doxygen-see-also` section.
+
+You must redocument the parameters if the template specialization redefines any parameters (e.g., if the generic parameter ``T`` becomes ``T*`` in the specialization) or if it places additional restrictions on their values.
+
```
ktlim: Add an example?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+.. _cpp-doxygen-related:
+
+Helper Functions
+----------------
+
+Some operations on a class, particularly arithmetic operations, must be implemented as standalone functions even though they are *conceptually* part of the class. These functions should have the ``@relatesalso`` tag, followed by the name of the appropriate class. They will appear on the class's documentation page under the heading "Related Functions". Use this tag sparingly.
+
+For internal consistency, always use ``@relatesalso`` and not the synonymous ``@relatedalso``.
+
```
ktlim: Add an example?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+.. _cpp-doxygen-initializer:
+
+Initializer Declaration
+-----------------------
+
+By default, Doxygen shows the values of constants unless they are very long. The ``@showinitializer`` and ``@hideinitializer`` tags override this behavior.
+
```
ktlim: Add an example?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+.. _cpp-doxygen-notes:
+
+Notes
+-----
+
+*Notes* is an optional section that provides additional information about the code, possibly including a discussion of the algorithm or known limitations of the code. The notes should be prefixed by a ``@note`` or ``@warning`` command.
+
+For internal consistency, always use ``@note`` and not the synonymous ``@remark`` or ``@remarks``.
+
```
ktlim: Add an example?  Also, perhaps cite the math documentation above in "Extended Summary"?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+   /**
+    * @addtogroup group
     *
-    * @brief Provide some stuff to do stuff
+    * @{
     */
 
-Class Definitions
-=================
+before any documented code, and a Doxygen comment with a ``@}`` after any documented code. Everything between the ``@{`` and the ``@}`` will be listed on the module page.
```
ktlim: We used to use `@ingroup`, which apparently didn't need bracketing.  Does that not work?  Requiring the `@}` at the end is unfortunate.

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+A ``@returns`` tag, followed by a description similar to the one for :ref:`cpp-doxygen-parameters`. If the returned value is a map, ensure that the key-value pairs are documented in the description.
+
+For consistency with Python documentation, always use ``@returns`` and not the synonymous ``@return``.
+
+.. _cpp-doxygen-throws:
+
+Throws
+------
+
+A series of ``@throws`` tags, one for each type of exception (see :ref:`the style guide <style-guide-cpp-5-36>`). Each tag should have a description following the exception type. If the description extends over multiple lines, each line after the first must be indented.
```
ktlim: Python's "Raises" section says "This section should be used judiciously—only for errors that are non-obvious or have a large chance of getting raised."  Something like that might be appropriate here; trying to get complete and exact exception documentation seems like it could easily become unbearable for both the writer and reader.

docs/cpp_docs.rst
```diff
@@ -128,77 +593,181 @@ For example:
...
+    * @exceptsafe Strong exception guarantee.
+    */
+   lsst::afw::image::Image<double> loadImage(std::string const & fileName);
+
+.. _cpp-doxygen-method-function-overloads:
+
+Overloaded Function/Method Definitions
+--------------------------------------
+
+``@overload`` may be used when two methods/functions are effectively the same but have different parameters list for reasons of convenience.
```
ktlim: "different parameter lists"?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
 .. code-block:: cpp
 
    /**
-    * @defgroup PackageName PackageTitle
+    * @defgroup afw LSST Framework
+    *
+    * @brief LSST data management: astronomical framework, including images and tables
+    */
+
+Header files that belong to the package must have a Doxygen comment like
```
ktlim: When this change is announced, please publicize that `@file` and `@author` are no longer present.

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+
+See Also
+--------
+
+'See Also' is an optional section used to refer to related code.
+This section can be very useful, but should be used judiciously.
+The goal is to direct users to other functions they may not be aware of, or have easy means of discovering (by looking at the class or package documentation, for example).
+Functions whose documentation further explains parameters used by this function are good candidates.
+
+List each class, function, or method using a ``@see`` tag:
```
ktlim: The Python documentation says how to refer to classes and functions in other namespaces; should something similar perhaps go here?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+.. _cpp-doxygen-extended-summary:
+
+Extended Summary
+----------------
+
+A few sentences giving an extended description.
+This section should be used to clarify *functionality*, not to discuss implementation detail or background theory, which should rather be explored in the :ref:`cpp-doxygen-notes` section below.
+You may refer to the parameters and the function name, but parameter descriptions still belong in the :ref:`cpp-doxygen-parameters` section unless they are very lengthy.
+
+This section may include mathematical equations to describe the behavior of a class or method, but be sure to put math that pertains only to the implementation in the :ref:`cpp-doxygen-notes` section rather than the extended description.
```
ktlim: Math is described in "Notes" in the Python guide rather than in "Extended Summary".  Is the thinking that it's more likely to be used for behavior and not implementation in C++ than in Python?

docs/cpp_docs.rst
```diff
@@ -26,94 +45,540 @@ The beginning of both header and source code files should include
...
+Note that Web pages should be referenced with regular inline links.
+
+References are meant to augment the documentation, but should not be required to understand it.
+
+.. _cpp-doxygen-examples:
+
+Examples
+--------
+
+'Examples' is an optional section for examples. This section is very strongly encouraged.
```
ktlim: Python says "These examples do not replace unit tests, but are intended to be tested to ensure documentation and code are consistent."  Should that be carried over?

docs/cpp_docs.rst
r-owen: If so, and if's an easy change, an example showing normal parameters might be a bit more helpful

jonathansick:
@kfindeisen As per a Slack conversation today in `#dm-docs`, could you revert the suggested legal boilerplate to what's conventionally used in the Stack (i.e., not point to the COPYRIGHT file)? I jumped the gun and put RFC-45-ish boilerplate in before it was ready to be there.

docs/cpp_docs.rst
kfindeisen: This was a change requested by @TallJimbo and discussed [here](https://github.com/lsst-dm/dm_dev_guide/pull/92#discussion_r93306370).

docs/cpp_docs.rst
kfindeisen: There's a normal parameter example in the following section. This was just a copy-and-paste error.

docs/cpp_docs.rst
kfindeisen: Doing so would contradict the C++ style guide (as linked). IMO this is a case where differences in the languages and in their respective programming communities justify taking a different policy for C++ and Python.

docs/cpp_docs.rst
kfindeisen: It's recommended rather than optional because exception safety is a key part of C++ programming, and it's something the writer of a function/method needs to think about (and in practice, basic exception safety is fairly easy to achieve, especially in the age of smart pointers). In the absence of a guarantee, downstream code cannot handle any exceptions that do get thrown, out of fear that the program state can't be cleaned up.

I understand that there will be cases, especially legacy code, where no guarantee is possible, but exception-safe code is an established best practice in C++ and should be encouraged.

docs/cpp_docs.rst
kfindeisen: While it is possible to use compileable examples with Doxygen by including snippets of external files, I haven't been able to think of a scheme to do so robustly -- for example, if the snippets are included using line numbers, any nontrivial change to the example source code will break the rendered version. So IMO this is something best left out or deferred to another ticket.

docs/cpp_docs.rst
kfindeisen: No, it often did not work. For example, [the afw module documentation](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/group__afw.html) is a list of source files rather than a list of API elements.

docs/cpp_docs.rst
ktlim: I agree that exception safety is relatively easy to achieve, but by the same token, I think it may be wasteful to have almost every function say "@exceptsafe basic".  I agree that there are some cases (especially your example where calling code is expected to catch and handle exceptions) where it makes sense to call it out.

Another point: even worse than unmarked exception-unsafe code is code that claims to be exception-safe but isn't, or was but now is not due to some change, perhaps in a distant package.

I think this may come down to judging whether exception handling (and throwing) is going to be common or rare.  I tend towards the latter.

docs/cpp_docs.rst
ktlim: Comparing with the style guide, I'm fine with having "@throws" for all exceptions actually thrown, but not all that might propagate up.  Since the latter isn't mentioned in the text, I think I was just over-interpreting and this can remain as is.

docs/cpp_docs.rst
ktlim: OK, that's fine, but we should consider making the Python version parallel, then.

docs/cpp_docs.rst
ktlim: Yup, I realize that this is a difficult problem.  No problem leaving it out.

docs/cpp_docs.rst
kfindeisen: It may be wasteful once developers can safely assume code is exception-safe, but is that something they can do now, or something the Stack needs to work towards?

However, you do have a good point that it depends on catching policy. A quick search of the Stack on GitHub shows many catch blocks don't do anything (or, at most, log a report), and many even just catch `Exception`. Perhaps exception safety is the least of such code's worries.  :disappointed:

A final consistency question: extended descriptions are listed as "recommended", though there are many functions for which it would be redundant. If I'm to demote exception safety to "optional", should I do the same for descriptions?

PR #94, opened 2016-12-22
Make code layout comaptible with clang-format (RFC-261)

pschella:


coding/cpp_style_guide.rst
```diff
@@ -1984,432 +1982,64 @@ We should, of course, strive to minimize the amount of such information.
 
 .. _style-guide-cpp-6-1:
 
-6-1. Multiple statements per line SHOULD NOT be used.
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. code-block:: cpp
+6-1. We follow the clang-format Google style for code layout with a few exceptions.
```
timj: Aren't we saying that we follow the Google style for code layout with a few exceptions (in which case a link to the Google style guide would be very useful)? clang-format is a tool that can be used to ensure that the style is followed. Or are we actually saying that clang-format with the supplied configuration file is the definition of our standard rather than a tool that can be used to enforce it?

pschella: I personally like having the tool define the style, because then there can be no discussion. But I agree that making "We follow the Google style guide for code layout with a few exceptions" the rule (with a link) perhaps makes more sense here. And then keep the remainder of the section on clang-format intact but list the exceptions more explicitly and change the emphasis.

PaulPrice: If this goes forward, we need a primer on what the new style is, and how it differs from the current style.

PaulPrice: We also need a eups build file for clang-format.

timj: @PaulPrice if you mean an EUPS package with `clang-format` in it then I'm not sure why we need that (it's in homebrew for example). Are you asking for a `scons` target or something that can make it easy for people to update their code?

PaulPrice:
I think that if we're mandating that the result looks like what you get by running a particular tool, we need to make that tool easily available.  Having it in Homebrew is all very well for my MacBook, but we need a way to install it as unprivileged users on our department-controlled Linux servers.  So that means we need a eups build file or instructions on how to install it as an unprivileged user for the most common package managers our developers use.

pschella:
So it looks like these problems will go away if we simply mandate the resulting layout to follow the rules in the formatting section of the Google style guide (with a 4 spaces 110 column exception if we want). The output of the tool will be automatically compatible so those that want to use it can, but it is not required.

The specific rules to follow will then be linked to as: https://google.github.io/styleguide/cppguide.html#Formatting

pschella:
Following the example from the Python style guide, I have separated the style from the tool section (which is now optional) and turned section 6 into a (small) delta on the Google C++ Style Guide formatting section.

pschella:
Thank you for the feedback. I'm now sending this in for review.

P.S. Given that use of clang-format is now entirely voluntary I don't think we have to provide a build.

PaulPrice:
I've just glanced at the Google C++ style, and am surprised at the amount of "optional" and "either ... or" statements.  If readability of C++ is such a matter of judgement, can we trust clang-format to get it right?
I'm also surprised to find that the formats are subtly different in ways that I think some of our C++ gurus don't like (I wonder if @TallJimbo likes [this](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions) and [this](https://google.github.io/styleguide/cppguide.html#Function_Calls)).
I'm very surprised that we have moved so quickly on a major change to a C++ style we have used successfully for many years.

PaulPrice:
Furthermore, it seems to me that the adoption of the Google C++ style has not been approved by RFC, only the use of the `clang-format` with an LSST style file.

TallJimbo:
I've mostly just bought in to @pschella's assertion that the advantage of not having to think about formatting outweighs any quibbles I might have with the actual style we adopt, and that I can trust the smart people at Google to have come up with a style good enough that I won't be bothered too much by it.

Actually changing our style was part of [RFC-261](https://jira.lsstcorp.org/browse/RFC-261), but I do share your concern that this (perhaps more important part) was a bit buried in the description and was not at all obvious from the title.

My biggest concern is that `clang-format` might leave some important things unformatted; that seemed to have been the case in some of the examples posted to c.l.o.  But I'm willing to see how that goes.  If it turns out we still have to do a lot of formatting ourselves, we'll at least have gotten our style closer to one of the major external ones, and that may be a win on its own.

TallJimbo:
I should add that I agree a bit with @PaulPrice that we should consider putting a bit of effort into making `clang-format` available to developers.  Mandating outcomes instead of tools is all well and good, but this outcome is advantageous specifically because it allows a tool and without that tool we're all stuck reading and learning a new style guide, which is not at all advantageous.  I'm not sure we need to have the same level of access as a third-party package that's needed to actually build the code (e.g. ability install without root access on Linux), but a section in the dev guide about how to get it on popular platforms and perhaps integrate it with common editors seems like a *very* good idea.

r-owen:
This looks fine to me, what I can see of it. I would appreciate a fully formatted version because it's hard to see what's still in the style guide without that.

@PaulPrice note that google provides an emacs settings file. That may help. Nonetheless I agree with you that we ought to provide a clang-format package (or at least try to do so).

I agree with @PaulPrice that it is a pity the google style guide has optional variants where it is purely a matter of taste, such as allowing space inside parenthesis but preferring not to have it. I hope `clang-format` will pick the preferred choice, and I think we should mandate its use.

I agree with @TallJimbo that I'll be disappointed if `clang-format` leaves important things unformatted. If that proves so (and I think a bit of testing is in order) then is that fixable using a different configuration or different style? That said, I like the google style. It seems sensible and close to what we use.

One final remark: I think we should point out explicitly that our version of the google style guide doubles the indentation in all cases (because many rules are written with explicit indentation).

pschella:
@PaulPrice with regards to "optional" in the style. `clang-format` picks only one of the options (sometimes falling back to another one if the line exceeds a certain length, per the rules in the style). I assume that the order of preference is set by the Google developers that implemented it.

@TallJimbo, @r-owen as for providing instructions to install `clang-format` or providing a build. I'm not sure how to do that. I installed `clang-format` on OSX through `brew` but I assume everyone has their own favourite way of doing that. It also doesn't make much sense to duplicate the information about installation and editor settings from the upstream `clang-format` documentation (linked in the dev-guide) I think.
But if we can agree on a list of platforms, editors and preferred installation methods I might compose a list of instructions. I don't think this particular section of the dev-guide is the right place for such instructions though. But again, I'm open to suggestions.

parejkoj:
As I see it was asked about above, you can view the formatted version of this ticket here: https://developer.lsst.io/v/DM-8713/

tools/clang_format.rst
```diff
@@ -0,0 +1,44 @@
...
+
+or through `Homebrew <http://brew.sh>`_ (on macOS):
+
+.. code-block:: bash
+
+    brew install clang-format
+
+For more platforms see `here <http://releases.llvm.org/download.html>`_.
+
+Place the following configuration file, named ``.clang-format``, in your top-level source directory:
```
ktlim: Are we/should we be recommending that this be added to git repos (and presumably `.gitignore`)?

tools/vim.rst
```diff
@@ -56,3 +56,24 @@ Explanation
...
+There is an integration for Vim which lets you run the ``clang-format`` standalone tool on your current buffer, optionally selecting regions to reformat. The integration has the form of a python-file which can be found under ``clang/tools/clang-format/clang-format.py``.
+
+This can be integrated by adding the following to your ``.vimrc``:
+
+.. code-block:: text
+
+  map <C-K> :pyf <path-to-this-file>/clang-format.py<cr>
+  imap <C-K> <c-o>:pyf <path-to-this-file>/clang-format.py<cr>
+
+The first line enables clang-format for NORMAL`` and VISUAL mode, the second line adds support for INSERT mode. Change ``C-K`` to another binding if you need clang-format on a different key (``C-K`` stands for ``Ctrl+k``).
```
ktlim: Looks like an extra set of backquotes here.

coding/cpp_style_guide.rst
```diff
@@ -767,7 +767,7 @@ Exception classes are really not part of the main design of the program, and nam
 
 .. code-block:: cpp
 
-   double & getElevation(unsigned int antennaId), void pointAntenna(Source const & source)
+   double &getElevation(unsigned int antennaId), void pointAntenna(Source const &source)
```
ktlim: Even if both are acceptable, I'd prefer that the examples attach the `&` to the type, rather than the function/variable.

coding/cpp_style_guide.rst
```diff
@@ -1624,11 +1623,13 @@ A programmer must never rely on this feature, since this might be confusing for
 
 .. code-block:: cpp
 
-   void MyDerivedClass::foo( double /* scalefactor */ ) { // OK
-       };
-     
-       void MyDerivedClass::foo( double ) { // OK
-       };
+   void MyDerivedClass::foo(double /* scalefactor */){
```
ktlim: Space before opening brace?

coding/cpp_style_guide.rst
```diff
@@ -1624,11 +1623,13 @@ A programmer must never rely on this feature, since this might be confusing for
...
+           // OK
+   };
+   
+   void MyDerivedClass::foo(double){
```
ktlim: Space before opening brace?

coding/cpp_style_guide.rst
```diff
@@ -2007,6 +2006,15 @@ Because ``override`` already implies ``virtual``, virtual should not be included
 6.1. Layout
 -----------
 
+.. _style-guide-cpp-6-0:
+
+6-0. Layout MAY be automated with clang-format.
```
ktlim: Add "with the LSST configuration"?

coding/cpp_style_guide.rst
```diff
@@ -2007,6 +2006,15 @@ Because ``override`` already implies ``virtual``, virtual should not be included
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The easiest way to comply with the :ref:`layout rules <style-guide-cpp-layout>` is to use `clang-format <http://clang.llvm.org/docs/ClangFormat.html>`_.
```
ktlim: Consider adding something like "The intent is that all of the layout rules are compatible with clang-format using the LSST configuration; if any conflicts arise, this rule ensures that the clang-format result is permitted to be used notwithstanding any other."

coding/cpp_style_guide.rst
```diff
@@ -2079,87 +2087,88 @@ Example 2 adds an additional line without significant increase in readability.
...
-   public:
-       SomeClass() {}
-   protected:
-       ...
-   private:
-       ...
-   };
-
-is permitted but not encouraged, and should not be used in new code.
+  - Any base class name should be on the same line as the subclass name,
```
ktlim: Make sure you note when publicizing this on e.g. community.lsst.org that this is a change from previously-permitted to required.

coding/cpp_style_guide.rst
```diff
@@ -2079,87 +2087,88 @@ Example 2 adds an additional line without significant increase in readability.
...
+    This rule is optional in small classes.
+  - Do not leave a blank line after these keywords.
+  - The ``public`` section should be first, followed by the ``protected`` and finally the ``private`` section.
 
 .. _style-guide-cpp-6-6:
 
 6-6. Function declarations MAY have any of the following three forms:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Return type on the same line as function name, parameters on the same line if they fit.
```
ktlim: Note when publicizing that there are changes in this area, particularly for opening brace.

coding/cpp_style_guide.rst
```diff
@@ -2245,28 +2265,28 @@ This follows from the general block rule above.
...
+          statements;
+          break;
+  
+      default:
+          statements;
+          break;
+  }
 
-Note that each ``case`` keyword is indented 2 spaces relative to the ``switch`` statement as a whole and the statement blocks are indented one indention (4 spaces).
+Note that each ``case`` keyword is indented 4 spaces relative to the ``switch`` statement as a whole and the statement blocks are also indented 4 spaces.
```
ktlim: Again, publicize changes here.

coding/cpp_style_guide.rst
```diff
@@ -2338,16 +2359,21 @@ This is general practice.
 6-16. The following white space conventions SHOULD be followed:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-- Conventional operators should be surrounded by a space character (except * and / and %)
+- Assignment operators always have spaces around them.
+- Other binary operators usually have spaces around them, but it's
```
ktlim: Publicize this change.

coding/cpp_style_guide.rst
```diff
@@ -2374,67 +2400,44 @@ However, this must be balanced with being able to see more of the code at a glan
...
-
-   AsciiFile  *file;
-   int        nPoints;
-   float      x, y;
+6-20.
+^^^^^
 
-Enhance readability.
-The variables are easier to spot from the types by alignment.
+*Deleted*
```
ktlim: I'm not sure people were following this, but again noting its removal is good.

coding/cpp_style_guide.rst
```diff
@@ -2374,67 +2400,44 @@ However, this must be balanced with being able to see more of the code at a glan
...
-       case PHASE_WATER : strcpy(string, "Water"); break;
-       case PHASE_GAS   : strcpy(string, "Gas");   break;
-   }
+6-21a.
+^^^^^^
 
-There are a number of places in the code where white space can be included to enhance readability even if this violates common guidelines.
-Many of these cases have to do with code alignment.
-General guidelines on code alignment are difficult to give, but the examples above should give a general clue. 
+*Deleted*
```
ktlim: Note this removal.

coding/cpp_style_guide.rst
```diff
@@ -2374,67 +2400,44 @@ However, this must be balanced with being able to see more of the code at a glan
...
 
 .. _style-guide-cpp-6-21b:
 
 6-21b. Nested namespaces SHOULD be aligned left with each level of nesting on a new line.
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The closing braces should all appear on one line and should be commented with the full nested namespace.
+The contents of namespaces are not indented.
```
ktlim: This may represent a change from current practice?

coding/cpp_style_guide.rst
```diff
@@ -2463,52 +2466,39 @@ In an international environment English is the preferred language.
...
-   This:
    /* Get an antenna reference
     * and if it is pointed at the Sun,
     * point it at the moon, after checking
     * to see if the moon is blue.
     */
-    
-   Antenna & = getAntenna("bima.ant1");
+   
+   Antenna &ant = getAntenna("bima.ant1");
```
ktlim: Again, I'd prefer `Antenna& ant`

coding/cpp_style_guide.rst
```diff
@@ -2463,52 +2466,39 @@ In an international environment English is the preferred language.
...
-   /* at the moon */                ant.pointAt(Planet.MOON);
-   /* after checking */        } else {
-   /* if the moon is blue */        ant.pointAt(Planet.VENUS);
-   }
-   Commenting in the second way makes code difficult to read and difficult to modify.
 
-.. _style-guide-cpp-6-25:
+This maintains consistency by positioning the comment at the same level as the code being discussed.
 
-6-25. Comments SHOULD be included relative to their position in the code.
```
ktlim: Did this inadvertently get deleted?  Or is it assumed to be merged into 6-24?

tools/clang_format.rst
pschella: I don't think we should do that. It is better to have just one at top level, and that is hard to mandate given that everyone has their own configuration.

tools/vim.rst
pschella: Indeed

coding/cpp_style_guide.rst
pschella: ok

coding/cpp_style_guide.rst
pschella: ok

coding/cpp_style_guide.rst
pschella: ok

coding/cpp_style_guide.rst
pschella: done

coding/cpp_style_guide.rst
pschella: ok

coding/cpp_style_guide.rst
pschella: Oops, restored.

tools/clang_format.rst
r-owen: I am not keen on one `.clang-format` file per repo because keeping them in synch sounds too difficult. Many in-editor tools allow the user to have a single config file, which is my personal preference. For users of `lsstsw` (which is likely a significant fraction of our software developers) we could also provide a .clang-format in the `build` directory.

r-owen:
One more general comment. I dislike some of the choices that `clang-format` makes when google style is enabled (such as seemingly random formatting of space around `&`), but I would still much rather have a tool that automatically formats and put up with its quirks than do manual formatting.

r-owen:
@pschella regarding installation instructions...Perhaps a separate linked page or wiki for this? I agree that suggesting `Homebrew` to install the tool on macOS is all we need to say on the subject.  I would be happy to provide instructions for Sublime Text and Atom on macOS. Both have some small quirks.

PR #95, opened 2017-01-06
Add note banning "if false" to python dev guide tickets/DM-7532

parejkoj:
This is the implementation of RFC-205.

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -783,3 +783,42 @@ Iterators and generators SHOULD be used to iterate over large data sets efficien
 -----------------------------------------------------------------------------------
 
 Use iterators, generators (classes that act like iterators) and generator expressions (expressions that act like iterators) to iterate over large data sets efficiently.
+
+.. _style-guide-py-disabled-code:
+
+``if False:`` and ``if True:`` SHOULD NOT be used
+-------------------------------------------------
+
+Code must not be placed inside ``if False:`` or ``if True:`` blocks, nor left commented out.
+Instead, debugging code and alternative implementations must be placed inside a "named" ``if`` statement.
+Such blocks should have a comment describing why they are disabled.
+They may have a comment describing the conditions under which said code can be removed (like the completion of a ticket or a particular date).
+For example, for code that will likely be removed in the future, once testing is completed:
+
+.. code-block:: py
+
+    # Delete old_thing() and the below "if" statement once all unittests are finished (DM-123456).
+    use_old_method = False
+    if use_old_method:
+        old_thing()
+    else:
+        new_thing()
+
+It is often beneficial to lift such debugging flags into the method's keyword arguments to allow users to decide which branch to run. For example:
+
+.. code-block:: py
+
+    def foo(x, debug_plots=False):
+        do_thing()
+        if debug_plots:
+            plot_thing()
+
+or, using ``lsstDebug``, which can be controlled as part of a command line task:
+
+.. code-block:: py
+
+    import lsstDebug
+    def foo(x):
+        do_thing()
+        if lsstDebug.Info(__name__).debug_plots:
+            plot_thing()


coding/python_style_guide.rst
```diff
@@ -783,3 +783,36 @@ Iterators and generators SHOULD be used to iterate over large data sets efficien
 -----------------------------------------------------------------------------------
 
 Use iterators, generators (classes that act like iterators) and generator expressions (expressions that act like iterators) to iterate over large data sets efficiently.
+
+`if False:` and `if True:` SHOULD NOT be used
```
jonathansick: Like the other sections, could you add a reference anchor before this section. Probably

```
.. _style-guide-py-feature-flags:
```

coding/python_style_guide.rst
```diff
@@ -783,3 +783,36 @@ Iterators and generators SHOULD be used to iterate over large data sets efficien
...
+---------------------------------------------
+
+Code must not be placed inside `if False:` or `if True:` blocks, nor commented out as a block-comment. For debugging code or if you are undecided about which particular implementation to use, such code must be placed inside a "named" `if` statement. Such blocks should have a comment describing why they are disabled. They may have a comment describing the conditions under which said code can be removed (e.g. completion of a ticket, a particular date). For example, for code that will likely be removed in the future, once testing is completed:
```
jonathansick: Like the other content here, I'd recommend *one sentence per line* formatting rather than *one paragraph per line formatting*. It makes diffs and GitHub PRs easier to do. But +💯 for not using hardwrap! :)

jonathansick: The inline code should have two backticks around them in reST, for example:

```
Code must not be placed inside ``if False:`` or ``if True:`` blocks
```

jonathansick: A comma after "e.g." is necessary, I think. Or just use plain English:

```
(like a completion of a ticket, or a particular date)
```

jonathansick: > nor commented out as a block-comment

The "as a block-comment" *might* be a bit redundant. Up to you. You could write

```
[...], nor left commented out.
```

jonathansick: > For debugging code or if you are undecided about which particular implementation to use, such code must be placed inside a "named" `if` statement.

There's an asymmetry in "debugging code" and "if you are undecided about which particular implementation to use."

Maybe try:

```
Instead, debugging code and alternative implementations must be placed inside a "named" `if` statement.
```

coding/python_style_guide.rst
```diff
@@ -783,3 +783,36 @@ Iterators and generators SHOULD be used to iterate over large data sets efficien
...
+It is often beneficial to lift such debugging flags into the method's keyword arguments to allow users to decide which branch to run. For example:
+
+.. code-block:: py
+
+    def foo(x, debug_plots=False):
+        do_thing()
+        if debug_plots:
+            plot_thing()
+
+or, using lsstDebug, which can be controlled as part of a commandline task:
```
jonathansick: Maybe put lsstDebug inside double backticks. Eventually this will be a link to the module's documentation page.

jonathansick: I think "commandline" is two words. I've seen it hyphenated too. I need to figure out our accepted style for this.

coding/python_style_guide.rst
parejkoj: Thanks. These are all very good changes.

coding/python_style_guide.rst
parejkoj: Right. I've named it "disabled-code".

coding/python_style_guide.rst
parejkoj: Ooof. Good point. I put in a space, but maybe "command-line-task" is the appropriate usage in our case, since that's a defined thing.

PR #102, opened 2017-02-01
clarify some verification cluster details

PaulPrice:
This adds a couple of bits that answer questions I had about the verification cluster when I started using it recently.

### DIFF
--- a/services/verification.rst
+++ b/services/verification.rst
@@ -211,6 +211,12 @@ One can observe that after the job resources have been granted, the user shell i
 the login node ``lsst-dev01``. The command ``srun`` can be utilized to run commands on the job's allocated 
 compute nodes. Commands issued without ``srun``  will still be executed locally on ``lsst-dev01``. 
 
+You can also use ``srun`` without first being allocated resources (via ``salloc``).
+For example, to immediately obtain a command-line prompt on a compute node: ::
+
+    % srun -I --pty bash
+
+
 SLURM Example Executing Tasks with Different Arguments
 ------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


services/verification.rst
```diff
@@ -20,11 +20,11 @@ This page is designed help you get started on the ``Verification Cluster``:
 Overview of the Verification Cluster
 ====================================
 
-``lsst-dev01`` is a system with 24 cores, 256 GB RAM, running the latest CentOS 7.x that serves as the front end of the 
+``lsst-dev01`` is a system with 24 physical cores, 256 GB RAM, running the latest CentOS 7.x that serves as the front end of the
 ``Verification Cluster``.  ``lsst-dev01`` is described in further detail at :doc:`lsst-dev` and on the
 page of `available development servers <https://confluence.lsstcorp.org/display/LDMDG/DM+Development+Servers>`_ .
 
-The ``Verification Cluster`` consists of 48  Dell C6320 nodes with 24 physical cores (2 sockets, 12 cores per processor) and 128 GB RAM.  As such, the system provides a total of 1152 cores. 
+The ``Verification Cluster`` consists of 48  Dell C6320 nodes with 24 physical cores (2 sockets, 12 cores per processor) and 128 GB RAM.  As such, the system provides a total of 1152 physical cores. Due to hyperthreading, each physical core appears as 2 Linux cores, so users can access 48 cores per node.
```
daues: Overall looks ok.  We early on did not choose to emphasize the hyper threading in the docs as 'typical lsst processing' (if that is meaningful) seemed to involve a lot of multiprocessing rather than use of lightweight threads.  As such we thought not to overly encourage, e.g.,  running 48 processCcd on a node (I think i tried it, and it was quite lousy.)   Could the last line emphasize potential limitations by saying 'users can access 48  logical or virtual cores per node' or something like that?  Obtaining benefits would certainly require that the user understands the character of their processing.

PR #111, opened 2017-02-17
DM-7240: Recommend use of standard unittest python asserts

timj:


coding/python_testing.rst
```diff
@@ -34,6 +34,10 @@ The important things to note in this example are:
 * Test classes must, ultimately, inherit from :class:`unittest.TestCase` in order to be discovered, and it is :ref:`recommended <py-utils-tests>` that :lclass:`lsst.utils.tests.TestCase` be used as the base class when :lmod:`~lsst.afw` objects are involved.
   The tests themselves must be methods of the test class with names that begin with ``test``.
   All other methods and classes will be ignored by the test system but can be used by tests.
+* Specific test asserts, such as :meth:`~unittest.TestCase.assertGreater`, :meth:`~unittest.TestCase.assertIsNone` or :meth:`~unittest.TestCase.assertIn`, should be used wherever possible.
+  It is always better to use a specific assert because the error message will contain more useful detail and the intent is more obvious to someone reading the code.
+  Only use :meth:`~unittest.TestCase.assertTrue` or :meth:`~unittest.TestCase.assertFalse` if you are checking a complex statement that is unsupported by other asserts.
+* When testing that an exception is raised always use :meth:`~unittest.TestCase.assertRaises` in conjunction with a :keyword:`with` statement.
```
parejkoj: I would rephrase this something like the following:
```
When testing that an exception is raised always use :meth:`~unittest.TestCase.assertRaises` as a context manager, for example `with self.assertRaises(TypeError): 1 + "2"`
```

timj: This sentence is specifically associated with the example code that includes that usage example so I don't think I need to duplicate that do I? How about I add line numbers to the example instead and reference the relevant line?

parejkoj: Yes, referring to the relevant line in the example would be fine. I would prefer it be worded as "as a context manager" instead of "in conjunction with with".

coding/python_testing.rst
```diff
@@ -34,6 +34,10 @@ The important things to note in this example are:
 * Test classes must, ultimately, inherit from :class:`unittest.TestCase` in order to be discovered, and it is :ref:`recommended <py-utils-tests>` that :lclass:`lsst.utils.tests.TestCase` be used as the base class when :lmod:`~lsst.afw` objects are involved.
   The tests themselves must be methods of the test class with names that begin with ``test``.
   All other methods and classes will be ignored by the test system but can be used by tests.
+* Specific test asserts, such as :meth:`~unittest.TestCase.assertGreater`, :meth:`~unittest.TestCase.assertIsNone` or :meth:`~unittest.TestCase.assertIn`, should be used wherever possible.
+  It is always better to use a specific assert because the error message will contain more useful detail and the intent is more obvious to someone reading the code.
+  Only use :meth:`~unittest.TestCase.assertTrue` or :meth:`~unittest.TestCase.assertFalse` if you are checking a complex statement that is unsupported by other asserts.
```
r-owen: I'd say "if you are checking a boolean, or a complex statement that is unsupported by other asserts.". We have plenty of boolean predicates that require testing.

timj: Ah yes, if the thing you are checking is actually a boolean then that is indeed okay :smile:

r-owen: Also, I hope that spelling out the full name (with library prefix) will not confuse users. I assume it is necessary in order for links to work, but they are used as `self.assertGreater` and so on. Still...the examples should make it clear.

timj: The prefix won't appear in the rendered version. The `~` causes the full path to be ignored but allows the links to work.

PR #112, opened 2017-02-21
Tickets/dm 9481

daues:


timj:
Can some of these commits be squashed? What's the merge from master onto the ticket branch doing?

daues:
The pull request #112 page was declaring that  it was necessary to merge changes that 
had been made on master to the ticket before proceeding, i.e., before  merging the dm 9481 changes to master.

timj:
That's github telling you there have been changes on master that can't be tested I think. The project development policy is to rebase branches to master rather than merge master to the branch. Continually merging master to the branch leads to a history that looks a bit confusing.

See https://developer.lsst.io/processes/workflow.html#review-preparation

This assumes that the developer guide git repository follows standard practice for DM (which I assumed it did).

daues:
The Squash & Merge option was attempted, and it looks like a single commit  b9267ac was merged.

timj:
Yes. It also doesn't have a merge commit. We generally don't use the Github merge button unless there are the specific circumstances of the branch commits being on the tip of master and all ready to merge. Rebasing has to happen on the client end first.

daues:
If the not-rebased-commits are already remote, then is it too late to 
follow the procedure 
...
git rebase -i master
# interactive rebase
git push --force
...

Or will all be well ?

timj:
You are always allowed to force push a ticket branch. You are not allowed to force push master.

PR #121, opened 2017-03-07
DM-7450: Update coding standard to reflect non-executable tests

mtpatter:
From RFC-215


ktlim:
While merging this would be fine (although I agree with Tim that the RFC generally should not be explicitly mentioned), it appears that the ticket is waiting for a significant addition from @jonathansick, so this will stay open.

ktlim:
The big reorg happened; should this PR be adjusted and merged?

jdswinbank:
Per discussion on DM-7450, I don't think we're expecting Maria's PR to be merged as-is, but rather for @jonathansick to take ownership of the ticket and come up with a modified version of this work. I'm therefore closing this down for now.

PR #124, opened 2017-03-08
DM-9087: Add guidelines on using properties on Python, as per RFC-279.

TallJimbo:


coding/python_style_guide.rst
```diff
@@ -822,3 +822,31 @@ or, using ``lsstDebug``, which can be controlled as part of a command line task:
...
+.. _style-guide-py-properties:
+
+Properties SHOULD be used when they behave like regular instance attributes
+---------------------------------------------------------------------------
+
+Properties SHOULD be added to Python objects to provide syntactic sugar for a getter (and possibly setter) when all of the following conditions are true:
+
+ - The getter object must return the same type the setter object accepts, or the types must have very similar interfaces (e.g. because they are part of the same class hierarchy, or they share an important common interface, such as a Python Sequence).
+
+ - Either the returned object must be immutable or modifying it must modify the parent object in the expected way. Note that it may be useful to have a getter return an immutable object (e.g. ``tuple`` instead of ``list``) to meet this criteria. This prevents confusing behavior in which ``a.b.c = v`` could be a silent no-op.
```
ktlim: "criteria" -> "criterion".  I don't think "parent object" is well-defined in this context; perhaps something above like "SHOULD be added to Python objects" -> "SHOULD be added to Python objects (which we will refer to as 'parents')"? Or "must modify the property-containing object"?

coding/python_style_guide.rst
```diff
@@ -822,3 +822,31 @@ or, using ``lsstDebug``, which can be controlled as part of a command line task:
...
+
+
+.. _style-guide-py-properties:
+
+Properties SHOULD be used when they behave like regular instance attributes
+---------------------------------------------------------------------------
+
+Properties SHOULD be added to Python objects to provide syntactic sugar for a getter (and possibly setter) when all of the following conditions are true:
+
+ - The getter object must return the same type the setter object accepts, or the types must have very similar interfaces (e.g. because they are part of the same class hierarchy, or they share an important common interface, such as a Python Sequence).
```
ktlim: Shouldn't this be "getter method" and "setter method"?

PR #130, opened 2017-03-20
Add caveats for devtoolset-3

ktlim:
After the pybind11 port, it appears that we cannot mix devtoolset-3 and base stacks.  Add this caveat to the document.

### DIFF
--- a/services/lsst-dev.rst
+++ b/services/lsst-dev.rst
@@ -116,8 +116,12 @@ Select Appropriate Developer Tools
 The ``lsst-dev01`` system is configured with the latest CentOS 7.x as its operating system.
 This release of CentOS provides an old set of development tools, centered around version 4.8.5 of the `GNU Compiler Collection`_ (GCC).
 While this version of GCC does satisfy the `prerequisites for building the LSST stack`_, it is still rather out of date.
-Therefore, you could enable the `Red Hat Developer Toolset`_ version 3 (``devtoolset-3``) which has been pre-installed.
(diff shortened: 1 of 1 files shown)


services/lsst-dev.rst
```diff
@@ -116,8 +116,11 @@ Select Appropriate Developer Tools
 The ``lsst-dev01`` system is configured with the latest CentOS 7.x as its operating system.
 This release of CentOS provides an old set of development tools, centered around version 4.8.5 of the `GNU Compiler Collection`_ (GCC).
 While this version of GCC does satisfy the `prerequisites for building the LSST stack`_, it is still rather out of date.
-Therefore, you could enable the `Red Hat Developer Toolset`_ version 3 (``devtoolset-3``) which has been pre-installed.
+However, this version of GCC is used for building the :ref:`shared stacks <lsst-dev-loadlsst>` available on the machine and currently must be used if building against those stacks.
+
+If you prefer, you could enable the `Red Hat Developer Toolset`_ version 3 (``devtoolset-3``) which has been pre-installed.
 This provides an updated toolchain, including GCC 4.9.2.
+If you enable this, you must build the entire stack using this toolchain.
```
jdswinbank: Consensus on Slack seems to be that this should be made more explicit: can we simply remove the mention of `devtoolset-3` altogether? Failing that, the wording should be strong: "you should not use the shared stack if you enable the devtoolset".

jdswinbank:
Any chance we could get this (or a variant on it) merged?

PR #131, opened 2017-03-24
DM-9919: Add flake8 path exclusion configuration

jonathansick:
This PR adds recommended flake8 linter path exclusion configuration to the Python Style Guide.

Draft: https://developer.lsst.io/v/DM-9919/coding/python_style_guide.html#flake8-configuration-files

## Justification

- `bin` is autogenerated, so it shouldn't be linted (whereas `bin.src/` *should* be linted).

- `doc/` includes `conf.py` for Sphinx which may not be readily lintable because of its global namespace usage for configuration.

- `__init__.py` modules shouldn't be linted since they're often just configuring a namespace, requiring 'unused' imports.

- `version.py` is autogenerated and shouldn't be linted.

- `tests/.tests` is autogenerated.

## Implementation note

I've implemented this commit as only showing the final recommended configuration. I don't think its necessary to justify each path exclusion, but we could iterate on this design decision.

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -136,15 +136,24 @@ The ignored error codes are :ref:`explained above <style-guide-py-ignored-errors
 Flake8 configuration files
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


jonathansick:
Added justification text in a way that doesn't get into the way too much.

PR #135, opened 2017-03-29
DM-7579: Update naming rules based on RFC-97

timj:
@jonathansick can you please take a quick look at this before I send for review from K-T? I've tried to put in the RFC-97 decision but my initial problem is that we are sometimes stating a PEP-8 exception that is not actually an exception so I've moved those up and left the exceptions lower down. I wonder if the N80x exceptions need their own sections now that they only apply to Science Pipelines and it may be better to have one item along the lines of "Science Pipeline shall use camelCase".

The variable naming one with `l`, `O` and the like is in PEP-8 so I've added a note saying that rather than letting there be an implication that this is against PEP-8.

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -527,17 +527,6 @@ Names which shadow a python built-in function may cause confusion for readers of
 Creating a more specific identifier is suggested to avoid collisions.
 In the case of *filter*, ``filter_name`` may be appropriate; for *filter objects*, something like ``filter_obj`` might be appropriate.
 
-.. _style-guide-py-naming-ext-modules:
-
-When a Python module wraps a C/C++ extension module, the C/C++ module SHOULD be named <module>Lib
--------------------------------------------------------------------------------------------------
-
-When an extension module written in C or C++ has an accompanying Python module that provides a higher level (e.g. more object oriented) interface, the C/C++ module should append ``Lib`` to the module's name (e.g. ``socketLib``).
-
-.. warning::
-
-  I think this rule has changed now that we use pybind11.
-
 .. _style-guide-py-naming-ambiguous:
 
 Names l (lowercase: el), O (uppercase: oh), I (uppercase: eye) MUST be avoided
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -75,16 +75,22 @@ E251
    Unexpected spaces around keyword / parameter equals.
    See :ref:`style-guide-py-multiline-assignment-whitespace`.
 
+Maximum line length
+   See :ref:`style-guide-py-line-length`.
+
+Additionally, packages listed in :ref:`style-guide-py-sci-pi-naming` should disable the following rules:
+
 N802
    Function name should be lowercase.
-   See :ref:`style-guide-py-naming`.
+   See :ref:`style-guide-py-sci-pi-naming`.
 
 N803
    Argument name should be lowercase.
-   See :ref:`style-guide-py-naming`.
+   See :ref:`style-guide-py-sci-pi-naming`.
 
-Maximum line length
-   See :ref:`style-guide-py-line-length`.
+N806
+   Variable in function should be lowercase.
+   See :ref:`style-guide-py-sci-pi-naming`.
 
 .. _pycodestyle error: http://pep8.readthedocs.io/en/latest/intro.html#error-codes
 
@@ -124,12 +130,13 @@ Flake8 command line invocation
 
 .. code-block:: bash
 
-   flake8 --ignore=E133,E226,E228,N802,N803 --max-line-length=110 .
+   flake8 --ignore=E133,E226,E228 --max-line-length=110 .
 
 This command lints all Python files in the current directory.
 Alternatively, individual files can be specified in place of ``.``.
 
 The ignored error codes are :ref:`explained above <style-guide-py-ignored-errors>`.
+N802, N803, and N806 can be added to this list for some packages.
 
 .. _style-guide-py-flake8-config:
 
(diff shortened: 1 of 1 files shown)


jonathansick:
Nice. I'll look at this by tomorrow morning and also re-assign the JIRA ticket to you.

coding/python_style_guide.rst
```diff
@@ -469,22 +469,22 @@ This comment should appear after the ``def`` line.
 We follow `PEP 8ʼs naming conventions <https://www.python.org/dev/peps/pep-0008/#naming-conventions>`_, with exceptions listed here.
 The naming conventions for LSST Python and C++ source have been defined to be as similar as the respective languages allow.
 
-In general:
+All LSST Python source code is consistent with PEP-8 naming as follows:
 
 - class names are ``CamelCase`` with leading uppercase,
```
jonathansick: It's my fault, but since you're here, you could convert each of these list items into  complete sentences (start with a capital, end with a full stop). Thanks!

coding/python_style_guide.rst
```diff
@@ -469,22 +469,22 @@ This comment should appear after the ``def`` line.
...
 - module variables used as module global constants are ``UPPERCASE_WITH_UNDERSCORES``,
-- all other names are ``camelCase`` with leading lowercase.
```
jonathansick: Oh I understand what happened to my earlier list. Here's the last item. Generally it's awkward to have a 1-item list.

Maybe splitting off Science Pipelines Naming Conventions into a following section would help here.

timj:
Thanks. I worry a bit that the anchors to the individual rules in PEP-8 are not right. 

For example, the "Class Attributes" rule links to https://www.python.org/dev/peps/pep-0008/#id45 which is a link to a table of contents subsection link for "Function and Method Arguments" ending up at https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments -- it's not clear to me how "Class Attributes" relates to "Always use `self` and `cls`...".  It looks like the layout is a bit weird as if you click on the section heading you go to the table of contents and then that takes you back to the section so all the `#idNN` links need to be fixed.

timj:
@jonathansick I've had a go at reorganizing. I've also removed the SWIG rule and fixed the PEP8 links (the `#idNN` links were wrong).

coding/python_style_guide.rst
```diff
@@ -467,40 +474,42 @@ This comment should appear after the ``def`` line.
 =====================
 
 We follow `PEP 8ʼs naming conventions <https://www.python.org/dev/peps/pep-0008/#naming-conventions>`_, with exceptions listed here.
-The naming conventions for LSST Python and C++ source have been defined to be as similar as the respective languages allow.
+C++ source code included within a Python package would normally follow the naming conventions of the Python package for APIs that are to be visible to Python users.
```
ktlim: "would normally" is not a good verb here.  Given the discoverability issues, I think this should be "SHOULD".

coding/python_style_guide.rst
```diff
@@ -467,40 +474,42 @@ This comment should appear after the ``def`` line.
...
-In general:
+All LSST Python source code is consistent with :pep:`8` naming in the following ways:
 
 - class names are ``CamelCase`` with leading uppercase,
 - module variables used as module global constants are ``UPPERCASE_WITH_UNDERSCORES``,
-- all other names are ``camelCase`` with leading lowercase.
+
+Some packages, for historical reasons, do not fully adhere to :pep:`8`.
+These packages, and the associated naming conventions, are described in :ref:`style-guide-py-sci-pi-naming`.
+Code in other packages SHOULD be consistent within a top-level package built by Jenkins, or within a distinct service, and it is RECOMMENDED that :pep:`8` naming convention be adopted for these, whilst understanding that it may be difficult to modify existing packages.
```
ktlim: Do you really need the "in other packages" here?

coding/python_style_guide.rst
```diff
@@ -467,40 +474,42 @@ This comment should appear after the ``def`` line.
...
+Consistency within a package is important.
```
ktlim: Is this strong enough?  Consistency between packages in the same product is important, consistency within the same package would seem mandatory.

coding/python_style_guide.rst
```diff
@@ -467,40 +474,42 @@ This comment should appear after the ``def`` line.
...
 
-User defined names SHOULD NOT shadow python built-in functions
---------------------------------------------------------------
+Naming Conventions for Science Pipelines
+----------------------------------------
 
-Names which shadow a python built-in function may cause confusion for readers of the code.
-Creating a more specific identifier is suggested to avoid collisions.
-In the case of *filter*, ``filterName`` may be appropriate; for *filter objects*, something like ``filterObj`` might be appropriate.
+For historical reasons, Science Pipelines code (nominally, all packages included in the ``lsst_apps`` metapackage, as well as ``meas_*``, ``pipe_*``, and ``obs_*`` and all dependencies), uses ``camelCase`` rather than completely adhering to :pep:`8`-style.
```
ktlim: `camelCase` is used only sometimes.  I'd change "uses camelCase rather than completely adhering" to "does not completely adhere".

coding/python_style_guide.rst
```diff
@@ -509,12 +518,14 @@ Modules which contain class definitions SHOULD be named after the class name
...
 
-When a Python module wraps a C/C++ extension module, the C/C++ module SHOULD be named <module>Lib
--------------------------------------------------------------------------------------------------
+User defined names SHOULD NOT shadow python built-in functions
+--------------------------------------------------------------
 
-When an extension module written in C or C++ has an accompanying Python module that provides a higher level (e.g. more object oriented) interface, the C/C++ module should append ``Lib`` to the module's name (e.g. ``socketLib``).
+Names which shadow a python built-in function may cause confusion for readers of the code.
+Creating a more specific identifier is suggested to avoid collisions.
+In the case of *filter*, ``filter_name`` may be appropriate; for *filter objects*, something like ``filter_obj`` might be appropriate.
```
ktlim: Maybe add a "For example," here (which wasn't in the original).

PR #136, opened 2017-04-06
DM-9596: __all__ = ... should go after from __future__ import and before any other includes

r-owen:


coding/python_style_guide.rst
```diff
@@ -582,15 +582,13 @@ Within a module, follow the order:
...
-6. Private module variables (names start with underscore)
-7. Private module functions and classes (names start with underscore)
-8. Public module variables
-9. Public functions and classes
-10. Optional test suites
-
-.. FIXME JSick: Delete mention of test suites?
+4. ``from __future__ import absolute_import, division, print_function``
+5. ``__all__ = [...]`` statement, if present
+6. Imports
```
ktlim: Do you think this is clear enough, or would it be better to say "Other imports" or something like that?

PR #145, opened 2017-05-25
DM-10606: Document use of LD_PROFILE

PaulPrice:


jonathansick:
https://developer.lsst.io/v/DM-10606/coding/profiling.html#c-profiling

PR #152, opened 2017-07-07
DM-11225: Improve Git LFS configuration docs for lsst-dev

jonathansick:
- https://developer.lsst.io/v/DM-11225/services/lsst-dev.html#configure-git-lfs (Using Git LFS on lsst-dev)
- https://developer.lsst.io/v/DM-11225/tools/git_lfs.html#basic-git-lfs-configuration (adds a config section on running `git lfs install`, which is useful for those who got git-lfs from EUPS rather than installing it themselves).

### DIFF
--- a/services/lsst-dev.rst
+++ b/services/lsst-dev.rst
@@ -14,6 +14,7 @@ This page is designed to assist developers in their work on ``lsst-dev01``:
 #. :ref:`lsst-dev-ssh-keys`
 #. :ref:`lsst-dev-tools`
 #. :ref:`lsst-dev-loadlsst`
+#. :ref:`lsst-dev-gitlfs`
 #. :ref:`lsst-dev-xpra`
 
 .. _lsst-dev-overview:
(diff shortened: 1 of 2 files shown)


services/lsst-dev.rst
```diff
@@ -209,6 +210,22 @@ Note that the SSD and GPFS-backed stacks are independent: while both will automa
...
+.. _lsst-dev-gitlfs:
+
+Configure Git LFS
+=================
+
+You can get Git LFS through EUPS:
+
+.. code-block:: bash
+
+   setup -j git_lfs
```
jdswinbank: I don't think you actually need the `-j` here (although it shouldn't do any harm).

tools/git_lfs.rst
```diff
@@ -34,6 +34,19 @@ Git LFS requires Git version 1.8.2 or later to be installed.
 
 Before you can use Git LFS with LSST data you'll need to configure by following the next section.
 
+.. _git-lfs-basic-config:
+
+Basic Git LFS configuration
```
jdswinbank: Should this section be listed under "On this page" above? The other section headings at this level are.

Actually, it might make more sense to have a single "configuring Git LFS" section with "basic" and "for LSST" as subsections.

jonathansick: I like the single "Configuring Git LFS" section with subsections. I'll go with that. :+1:

PR #156, opened 2017-07-31
First draft of how to write LDM doc in Latex

timj:


docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
+###################################
+Writing Change-Controlled Documents
+###################################
+
+Change-controlled documents are documents that require approval from the Change Control Board (CCB) before they can be released.
+The project-wide change control process is described in `LPM-19`_, with the process for managing documents being described in `LPM-51`_, and covers those documents using DocuShare handles "LSE" or "LPM", that must be approved by the Project `CCB`_.
```
ktlim: No comma before "that".

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
+###################################
+Writing Change-Controlled Documents
+###################################
+
+Change-controlled documents are documents that require approval from the Change Control Board (CCB) before they can be released.
```
ktlim: "the" ➡️ "a"

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+The project-wide change control process is described in `LPM-19`_, with the process for managing documents being described in `LPM-51`_, and covers those documents using DocuShare handles "LSE" or "LPM", that must be approved by the Project `CCB`_.
+For Data Management, these documents have the DocuShare handle "LDM" and they must be approved by the DM CCB using the :ref:`RFC <decision-making-rfc>` process.
```
ktlim: Comma before "and".

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+These change-controlled documents can be written using the Word templates defined in `Document-9224`_, but the preference within Data Management is to write change-controlled documents in LaTeX using the ``lsstdoc`` document class provided by the `lsst-texmf`_ package, and to develop these documents using ``git`` repositories within the LSST organization on `GitHub`_.
```
ktlim: No comma before "and".

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+.. _CCB: https://project.lsst.org/groups/ccb/
+.. _Document-9224: https://ls.st/Document-9224
+.. _lsst-texmf: https://lsst-texmf.lsst.io
+.. _LPM-19: https://ls.st/LPM-19
+.. _LPM-51: https://ls.st/LPM-51
+.. _GitHub: https://github.com/lsst
+.. _LDM-294: https://ls.st/LDM-294
+
+.. note::
+  The definition of which types of documents should be change-controlled is defined in `LDM-294`_.
```
ktlim: "definition ... is defined" is redundant.  "defined" ➡️ "located"?

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+
+
+Developer Workflow
+==================
+
+The development of change-controlled documents differs from :ref:`that of code <git-branching>`, in that some documents under development can take a long time to evolve and can be developed under multiple stories (ticket branches), and by multiple authors.
+Repositories associated with change-controlled documents should therefore use two protected branches: ``master`` should correspond solely to an archived instance of the document on DocuShare, and ``draft`` is a development branch that should be used as the document evolves.
```
ktlim: "a development branch ... should" ➡️ "a long-lived integration branch ... can".

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+Ticket branches should branch from ``draft`` and should follow standard :ref:`DM branching policies <git-branching>`.
```
ktlim: To deal with the multiple-RFC problem in the "warning" below, I suggest that specific RFC-triggered updates go into `tickets/DM-NNNN` branches from `master`; only long-lived, multi-author, pre-RFC work would go into `draft`.  Hopefully there aren't two instances of the latter at the same time.

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+
+.. warning::
+  Need to understand whether we need to handle two independent edits to an LDM that will be submitted as two distinct RFCs in parallel.
+
+Once a document handle has been issued a repository can be created in the ``lsst`` GitHub organization using a repository name that matches the handle.
```
ktlim: Shouldn't we say that `cookiecutter` or the `sqrbot` equivalent should be used to ensure consistency and correctness and minimize follow-up credential work for auto-publishing?

docs/change_controlled.rst
```diff
@@ -0,0 +1,78 @@
...
+.. warning::
+  It may be necessary to adjust the history of a pre-existing repository if it has been developed using different policies, in particular it is important that ``master`` be adjusted such that it only contains merge commits from approved versions (and the initial placeholder commit).
+
+All documents use a Travis configuration file to build the document and deploy it to the corresponding https://ldm-nnn.lsst.io/ web site.
+Built PDFs of documents should not be committed to the repository.
+They are deployed automatically to the web during development and the final version will also be archived on DocuShare.
+
+During development:
+
+* Use the ``lsstdraft`` class option to make it clear that the document has not been approved.
```
ktlim: Can we auto-check in Travis to make sure that `lsstdraft` never appears in `master`?  (I don't think we can require that it appear in non-`master`, although that would be desirable.)

timj:
Thanks @ktlim -- I should have mentioned that this was a version from May, and that it will have to be tweaked following our experiences with the reality of the recent review. We do need to make a stand on submodules (preferring them but not insisting on them might be a good compromise).

docs/change_controlled.rst
timj: Yes. But currently that isn't done. We can update this once `sqrbot` can handle an LDM.

docs/change_controlled.rst
timj: I suppose that Travis could check although by that time you've already merged it to master.

timj:
I've made some tweaks based on feedback from @ktlim.

timj:
@ktlim should this page be change-controlled? I think it probably should be.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
+###################################
+Writing Change-Controlled Documents
+###################################
+
+Change-controlled documents are documents that require approval from a Change Control Board (CCB) before they can be released.
+The project-wide change control process is described in `LPM-19`_, with the process for managing documents being described in `LPM-51`_, and covers those documents using DocuShare handles "LSE" or "LPM" that must be approved by the Project `CCB`_.
+For Data Management, these documents have the DocuShare handle "LDM", and they must be approved by the DM CCB using the :ref:`RFC <decision-making-rfc>` process.
+These change-controlled documents can be written using the Word templates defined in `Document-9224`_, but the preference within Data Management is to write change-controlled documents in LaTeX using the ``lsstdoc`` document class provided by the `lsst-texmf`_ package and to develop these documents using ``git`` repositories within the LSST organization on `GitHub`_.
```
jdswinbank: I _personally_ have a preference for LaTeX over Word, but this is the first time I recall seeing it expressed in official DM documentation — my understanding from e.g. LDM-294 is that the latter is a first class citizen. Do we really want to change that in this document?

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
+###################################
+Writing Change-Controlled Documents
+###################################
+
+Change-controlled documents are documents that require approval from a Change Control Board (CCB) before they can be released.
+The project-wide change control process is described in `LPM-19`_, with the process for managing documents being described in `LPM-51`_, and covers those documents using DocuShare handles "LSE" or "LPM" that must be approved by the Project `CCB`_.
+For Data Management, these documents have the DocuShare handle "LDM", and they must be approved by the DM CCB using the :ref:`RFC <decision-making-rfc>` process.
```
jdswinbank: Being picky, LDM-294 refers to the "DMCCB", rather than "DM CCB".

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. note::
+  The definition of which types of documents should be change-controlled is located in `LDM-294`_.
+
+
+Developer Workflow
+==================
+
+The development of change-controlled documents differs from :ref:`that of code <git-branching>`, in that some documents under development can take a long time to evolve and can be developed under multiple stories (ticket branches), and by multiple authors.
+Repositories associated with change-controlled documents should therefore use two protected branches: ``master`` should correspond solely to an archived and approved instance of the document on DocuShare, and ``draft`` is a long-lived integration branch that can be used as the document evolves.
```
jdswinbank: So it won't surprise you to learn that I'm unhappy with this: it reverses the usual git semantics, in which work-in-progress lives on `master` and released versions are expressed with tags or branches, without any obvious justification.

I'm aware that a bunch of people seem to like this so I won't push back too hard if this is what the majority really want, but I would appreciate it if you can explain the thinking behind this weird role-reversal — at least to me personally, but it'd be even better if such an explanation can be included in this document.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+
+There can be cases where some minor fixes to a controlled document must be made whilst more major changes are being developed using the ``draft`` branch.
+It is therefore allowed for ticket branches to be created off of ``master`` if the changes are shortly going to be sent for CCB approval.
```
jdswinbank: http://grammarist.com/usage/off-of/

Not a serious issue; it just grates a little.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+Given this requirement, it is preferred that ``draft`` be used in the usual case.
+Regardless of which base branch is used, ticket branches should follow standard :ref:`DM branching policies <git-branching>`.
+
+Once a document handle has been issued a repository can be created in the ``lsst`` GitHub organization using a repository name that matches the handle.
+If the document exists in a pre-existing repository, it should be renamed to the handle and, if necessary, moved to the correct organization.
+
+.. warning::
+  It may be necessary to adjust the history of a pre-existing repository if it has been developed using different policies, in particular it is important that ``master`` be adjusted such that it only contains merge commits from approved versions (and the initial placeholder commit).
+
+All documents use a Travis configuration file to build the document and deploy it to the corresponding https://ldm-nnn.lsst.io/ web site.
```
jdswinbank: Is this actually true? It certainly isn't at the moment.

Given that the introduction provides for this process covering documents which aren't simply LDMs, then the URL should probably be made more generic. Or are you suggesting that that this process only applies to LDM-type documents?

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+Built PDFs of documents should not be committed to the repository.
+They are deployed automatically to the web during development and the final version will also be archived on DocuShare.
```
jdswinbank: "Baselined", rather than "final", I guess, since baselined documents can be changed: nothing is ever finally final. Until it is.

docs/change_controlled.rst
jdswinbank: So... *which* of "these change-controlled documents can be written..." etc? Is the process defined in this document designed to be used for all of LPM, LSE, LDM, and others, or just for DM-level change control? Will we publish LPMs at `lpm-xxx.lsst.io`, for example?

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+
+During development:
+
+* Use the ``lsstdraft`` class option to make it clear that the document has not been approved.
+* Specify a date using :code:`\today` so that the built document continues to have an updated date for each change (we do not automatically insert a date based on the state of the git repository).
```
jdswinbank: Shouldn't we do this based on git? Naïvely, it seems like the obviously-correct approach.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+* Update the change record but do not specify a version number.
+  The version number will be added during the release process.
```
jdswinbank: "During development... the version number will be added during the release process". So it's not really during development, then? :-)

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+* New references should be added to a local bib file and that file should be added to the bibliography search path.
+* If the document is receiving minor edits consider using the :code:`\newtext` and :code:`\oldtext` macros to indicate text changes.
+  For larger changes the :command:`latexdiff-vc` command may be helpful when preparing for an RFC or change request submission.
+
+.. note::
+  Should draft mode automatically override the release date?
+
+When a document is ready to be reviewed by the CCB, it should be uploaded to DocuShare.
```
jdswinbank: Which CCB? DM or LSST? Or both?

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+The review process is more efficient if the changes to be reviewed are marked clearly on the PDF.
+If the CCB process triggers new modifications these should be made on a branch named ``tickets/RFC-nnn`` or ``tickets/LCR-nnn`` (relative to ``draft``) to make clear that the work was triggered by the review process.
```
jdswinbank: Why isn't this work ticketed in the normal way (ie, `DM-NNNN`)? It seems like something we should track through our regular process.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+Once the document change is approved the CCB shall nominate someone to prepare the document for release.
+The release process for a document consists of (starting on the LCR/RFC ticket branch):
+
+* Remove ``lsstdraft`` modifier and set the date to the approval date.
+
+.. note::
+  If you use the ``\XXX`` command or ``draftnote`` environment, these notes will be hidden automatically when draft mode is removed from the document.
+
+* Move references to the shared bibliography files in the ``lsst-texmf`` package.
```
jdswinbank: So, just to be clear — is the point that references should *not* be added to ``lsst-texmf`` until they are required for a document being baselined? If so, I think that's a new policy we should call out more explicitly.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+  References should be added in the correct files and should be placed in the file based on alphabetical ordering of the keys.
+  There should be no local bib file in a released document.
+* Update the change record following the process described in `LPM-51`_.
+  For LSE documents the change record should be approved by the LSST Librarian.
```
jdswinbank: I am assuming that the "LSST Librarian" is the individual referred to by LPM-51 as the "LSST Project Librarian". I'm also assuming it's Rob McKercher, but I can't find that written down anywhere obvious. Better not to hard code that name in this document, since it might change, but citing some canonical source would be good.

docs/change_controlled.rst
```diff
@@ -0,0 +1,82 @@
...
+* Verify that the generated PDF (preferably one built and deployed by Travis) looks correct.
+  LSE documents should be approved by the LSST Librarian at this point to ensure that the document is correct before the changes are merged to ``master``.
+  The LSST Librarian will upload the document to DocuShare and close out the LCR.
+* Merge the ticket branch to ``draft`` and the ``draft`` branch to master.
+* For DM change-controlled documents upload the PDF to DocuShare and make it the preferred version.
```
jdswinbank: It's a little confusing here to distinguish which steps refer to DM level change control, and which to project level. Maybe we need two separate checklists rather than trying to merge them?

(Probably doesn't help that you are — I think! — using "LSE" to refer to all Project-level change controlled documents, which might actually be LPMs or whatever.)

timj:
This work was done in #169.

PR #159, opened 2017-08-09
lsst/lsstsw -> lsst/repos split (repos.yaml)

jhoblitt:


jhoblitt:
I thought about it -- I was concerned that it would be more confusing than helpful.  `repos.yaml` will still appear at the same path relative to the root of the `lsstsw` clone, it just isn't checked into the repo any more.  I'll hold off on merging this and take a fresh look at it this afternoon.

jhoblitt:
Added a tiny bullet point to `build-ci/lsstsw.rst` and repushed.

PR #161, opened 2017-09-13
Update the pytest instructions

timj:


coding/python_testing.rst
```diff
@@ -9,8 +9,7 @@ Python Unit Testing
 
 This page provides technical guidance to developers writing unit tests for DM's Python code base.
 See :doc:`unit_test_policy` for an overview of LSST Stack testing.
-If you have legacy DM :mod:`unittest` ``suite``-based code (code that sets up a :class:`unittest.TestSuite` object by listing specific test classes and that uses :lfunc:`lsst.utils.tests.run` rather than :func:`unittest.main`), please refer to tech note `SQR-012`_ for porting instructions.
-LSST tests should be written using the :mod:`unittest` framework, with default test discovery, and should support being run using the `pytest`_ test runner.
+LSST tests should be written using the :mod:`unittest` framework, with default test discovery, and should support being run using the `pytest`_ test runner and from the command line.
```
ktlim: Maybe "as well as from the command line" would be a slight bit clearer.

coding/python_testing.rst
```diff
@@ -29,7 +28,7 @@ A simple :mod:`unittest` example is shown below:
 
 The important things to note in this example are:
 
-* Test file names must begin with ``test_`` for maximum compatibility with `pytest`_.
+* Test file names must begin with ``test_`` to allow `pytest`_ to automatically detect them without requiring an explicit list of tests be supplied.
```
ktlim: "without requiring an explicit test list, which can be hard to maintain and can lead to missed tests." might provide better rationale.

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
-=================================
+Supporting Pytest
+=================
 
 .. note::
-   `pytest`_ is not currently installed as part of a LSST stack installation.
-   It can be installed using :command:`conda install pytest` or :command:`pip install pytest`.
+  `pytest`_ and its plugins are standard stack EUPS packages and do not have to be installed separately.
+
+All LSST products that are built using :command:`scons` will execute Python tests using `pytest`_ and all tests should be written such that they are runnable using `pytest`_, a policy adopted in :jira:`RFC-69`.
```
ktlim: "and all tests should be ..." -> ", so all tests must be runnable using it."  We're trying to avoid mentioning RFCs directly in text.

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
+The behavior of `pytest`_ when invoked by :command:`scons` is controlled by the :file:`tests/SConscript` file.
+At its simplest, this file should contain the following to enable testing:
+
+.. code-block:: python
+
+   from lsst.sconsUtils import scripts
+   scripts.BasicSConscript.tests()
+
+When the ``tests`` target is executed the Python test files will be located with a glob for ``*.py`` in the :file:`tests` directory and the tests will be run with a single invocation of :command:`pytest` using the number of subprocesses matching the ``-j`` argument given to :command:`scons`.
+This is the classical method for test discovery but the preferred approach now is for packages to support automatic test discovery using `pytest`_ as described below.
```
ktlim: List the preferred approach first.

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
+
+``pyList`` is used to specify which Python test files to run.
+Here the empty list is interpreted as "allow :command:`pytest` to automatically discover tests", whereas, for backwards compatibility reasons, the default of ``pyList=None`` is interpreted as "allow :lmeth:`~lsst.sconsUtils.scripts.BasicSConscript.tests` to automatically discover tests", as described above.
+For this mode to work, all test files must be named :file:`test_*.py`.
+
+``pySingles`` is an optional argument that can be used for the rare cases where a test must be run standalone and not with other test files in a shared process.
+The tests are still run using :command:`pytest` but executed one at a time without using any multi-process execution.
+Use of this should be extremely rare.
+In the :lmod:`~lsst.base` package one test file is used to confirm that the LSST import code is working; this can only be tested if we know that it hasn't previously been imported as part of another test.
+The other reason, so far, to run a test standalone is for test classes that dynamically generate large amounts of test data in setup.
```
ktlim: "setup" or "setUp"?

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
+The convention is to name these files :file:`test*.py` without the underscore.
+
 .. note::
-   The normal way to run tests is to use :command:`scons` or :command:`scons tests`.
-   :command:`scons` will then execute each of the tests and report any failures.
-   Currently tests are executed one at a time using :command:`python` and checking the command exit status.
-   Eventually :lmod:`~lsst.sconsUtils` will be modified to call :command:`py.test` directly with all Python test files (possibly with a different order each invocation).
-   It is therefore important that during this transition period developers check that tests run correctly with `pytest`_ by explicitly running :command:`py.test` when creating or updating test files.
+  When :command:`pytest` is run by :command:`scons` full warnings are reported, including :py:class:`~exceptions.DeprecationWarning`.
+  Usually these warnings are hidden in the test output but now they are more obvious, allowing you to fix any problems early.
```
ktlim: Leave out the "Usually" part, since that is no longer usual.

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
+
+Where does the output go?
+=========================
+
+When :command:`scons` runs any tests, the output from those tests is written to the :file:`tests/.tests` directory, and a file is created for each test that is executed.
+If a test command fails, that output is renamed to have a :file:`.failed` extension and is reported by :command:`scons`.
+For the case where :command:`pytest` is running multiple of the test files, there is a single file created in the :file:`tests/.tests` directory, along with an XML file containing the test output in JUnit format.
```
ktlim: "multiple of the" -> "multiple".  Again, perhaps describe the normal situation before the unusual one.

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
+For convenience the output from the main :command:`pytest` run (as opposed to the rare standalone usages) is also written to standard output so is visible in the log or in the shell along with other :command:`scons` output.
```
ktlim: "so is" -> "so it is".

coding/python_testing.rst
timj: Yes. That text was there previously. I'll fix.

coding/python_testing.rst
timj: "the set up phase" (since it's most like `setUpClass`).

coding/python_testing.rst
timj: "Previously these warnings were hidden"?

ktlim: That would be fine.

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
+
+Common Issues
+=============
```
jonathansick: I recommend having a sub header for each issue, something like:

- Testing state by running tests in different orders
- Use the unittest skipping framework

coding/python_testing.rst
```diff
@@ -42,44 +41,134 @@ The important things to note in this example are:
...
 :mod:`unittest` supports skipping of individual tests and entire classes using decorators or skip exceptions.
 LSST code sometimes raises skip exceptions in :meth:`~unittest.TestCase.setUp` or :meth:`~unittest.TestCase.setUpClass` class methods.
 It is also possible to indicate that a particular test is expected to fail, being reported as an error if the test unexpectedly passes.
 Expected failures can be used to write test code that triggers a reported bug before the fix to the bug has been implemented and without causing the continuous integration system to die.
-One of the primary advantages of using a modern test runner such as `pytest`_ is that it is very easy to generate machine-readable pass/fail/skip/xfail statistics to see how the system is evolving over time.
+One of the primary advantages of using a modern test runner such as `pytest`_ is that it is very easy to generate machine-readable pass/fail/skip/xfail statistics to see how the system is evolving over time, and it is also easy to enable code coverage.
+Jenkins now provides test result information.
```
jonathansick: Reminds me that we need docs (another ticket) for how to use new Jenkins and get results...

timj:
I've pushed a new version. @ktlim do you want to sign off on this version?

coding/python_testing.rst
timj: Not helped by Blue Ocean hiding all the output except the failures and a note of the skipped tests (but not telling you the reason for the skip).

coding/python_testing.rst
```diff
@@ -42,44 +41,149 @@ The important things to note in this example are:
...
+At minimum this file should contain the following to enable testing with automated test discovery:
+
+.. code-block:: python
+
+   from lsst.sconsUtils import scripts
+   scripts.BasicSConscript.tests(pyList=[])
+
+``pyList`` is used to specify which Python test files to run.
+Here the empty list is interpreted as "allow :command:`pytest` to automatically discover tests"
+:command:`pytest` will scan the directory tree itself to find tests and will run them all in together using the number of subprocesses matching the ``-j`` argument given to :command:`scons`.
```
ktlim: "all in together" is an unusual usage to this American.  Perhaps just "all together"?

coding/python_testing.rst
```diff
@@ -29,7 +28,7 @@ A simple :mod:`unittest` example is shown below:
 
 The important things to note in this example are:
 
-* Test file names must begin with ``test_`` for maximum compatibility with `pytest`_.
+* Test file names must begin with ``test_`` to allow `pytest`_ to automatically detect them without  requiring an explicit test list, which can be hard to maintain and can lead to missed tests.
```
ktlim: Two spaces after "without".

coding/python_testing.rst
```diff
@@ -42,44 +41,149 @@ The important things to note in this example are:
...
+   scripts.BasicSConscript.tests(pyList=[], pySingles=["testSingle.py"])
+
+The tests are still run using :command:`pytest` but executed one at a time without using any multi-process execution.
+Use of this should be extremely rare.
+In the :lmod:`~lsst.base` package one test file is used to confirm that the LSST import code is working; this can only be tested if we know that it hasn't previously been imported as part of another test.
+The other reason, so far, to run a test standalone is for test classes that dynamically generate large amounts of test data during the set up phase.
+Until it is possible to pin test classes to a particular process with ``pytest-xdist``, tests such as these interact badly when test methods within the class are allocated to different subprocesses since each subprocess will generate the test files.
+This can use significantly more disk and CPU when the test runs, and can even cause Jenkins to fail.
+It is important to ensure that any files listed in ``pySingles`` should be named such that they will not be discovered by `pytest`_.
+The convention is to name these files :file:`test*.py` without the underscore.
```
ktlim: For my own edification, does `pySingles` interact properly with `pyList=None`?

coding/python_testing.rst
```diff
@@ -42,44 +41,149 @@ The important things to note in this example are:
...
+
 .. note::
-   The normal way to run tests is to use :command:`scons` or :command:`scons tests`.
-   :command:`scons` will then execute each of the tests and report any failures.
-   Currently tests are executed one at a time using :command:`python` and checking the command exit status.
-   Eventually :lmod:`~lsst.sconsUtils` will be modified to call :command:`py.test` directly with all Python test files (possibly with a different order each invocation).
-   It is therefore important that during this transition period developers check that tests run correctly with `pytest`_ by explicitly running :command:`py.test` when creating or updating test files.
+  When :command:`pytest` is run by :command:`scons` full warnings are reported, including :py:class:`~exceptions.DeprecationWarning`.
```
ktlim: This appears to be less-indented than the previous text.  Is that OK, should it be indented more, or should the `.. note::` be removed?

coding/python_testing.rst
```diff
@@ -42,44 +41,149 @@ The important things to note in this example are:
...
+  Previously these warnings were hidden in the test output but now they are more obvious, allowing you to fix any problems early.
+
+Where does the output go?
+-------------------------
+
+When :command:`scons` runs any tests, the output from those tests is written to the :file:`tests/.tests` directory, and a file is created for each test that is executed.
+If a test command fails, that output is renamed to have a :file:`.failed` extension and is reported by :command:`scons`.
+For the case where :command:`pytest` is running on multiple test files at once, there is a single file created in the :file:`tests/.tests` directory, along with an XML file containing the test output in JUnit format.
```
ktlim: I thought I had asked to move this case (where `pytest` is run on multiple test files) to be first, since it is the normal case now.  Also, is a `.failed` file created in this case?

coding/python_testing.rst
timj: I agree. I was in the middle of saying "all in one process" and then realized that `xdist` means that is clearly not true.

coding/python_testing.rst
timj: Of course :-)  `daf_base` is already doing that.

coding/python_testing.rst
timj: I think I'm happy with it being a note but I realize that it's now in the wrong place because of the new subsections. I'll move it up.

coding/python_testing.rst
timj: Yes. Sorry. I lost that in the big reorg that @jonathansick asked for. A `.failed` file is always created on failure still, even with the XML file. That was the easiest way to migrate everything.

PR #173, opened 2017-10-09
DM-10342: Switch to C++14

pschella:
Implements RFC-332

### DIFF
--- a/coding/cpp_style_guide.rst
+++ b/coding/cpp_style_guide.rst
@@ -84,10 +84,29 @@ These tasks are made much simpler if the code is easily readable and well-docume
 
 .. _style-guide-cpp-2-2:
 
-2-2. We are writing C++11/14 with some restrictions.
-----------------------------------------------------
+2-2. We are writing C++14
+-------------------------
 
-The official policy on the use of C++11 features is at :ref:`Policy on use of C++11/14 language features <style-guide-cpp-cpp-11-14>`.
+The C++11 standard and the C++14 improvements to it bring a number of useful language features that make the resulting code more expressive, easier to read, and safer.
+
+We follow the official: International Standard ISO/IEC 14882:2014(E) – Programming Language C++, without any compiler specific extensions.
+
+.. note::
+
+    Our minimum required compiler versions are:
+
+        * GCC 6.3.1 (Linux)
+        * Clang 800.0.42.1 (macOS)
+
+    these both have complete support for C++14, but in case of compiler bugs the actually allowed set
+    of C++14 features is the intersection of those supported by our compilers and the standard.
+
+.. seealso::
+
+   - :ref:`pipelines:source-install-redhat-legacy` from the `LSST Science Pipelines <https://pipelines.lsst.io>`__ documentation.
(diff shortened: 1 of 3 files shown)


coding/cpp_style_guide.rst
```diff
@@ -84,10 +84,17 @@ These tasks are made much simpler if the code is easily readable and well-docume
...
+-------------------------
 
-The official policy on the use of C++11 features is at :ref:`Policy on use of C++11/14 language features <style-guide-cpp-cpp-11-14>`.
+The C++11 standard and the C++14 improvements to it bring a number of useful language features that make the resulting code more expressive, easier to read, and safer.
+
+We follow the official: International Standard ISO/IEC 14882:2014(E) – Programming Language C++, without any compiler specific extensions.
+
+.. seealso::
+
+   - :ref:`pipelines:source-install-redhat-legacy` from the `LSST Science Pipelines <https://pipelines.lsst.io>`__ documentation.
```
ktlim: While this used to be sufficient, the docs for 14.0 that specify a minimum of gcc 4.8 are _not_ sufficient now.  We should make that clear.

coding/cpp_style_guide.rst
```diff
@@ -84,10 +84,17 @@ These tasks are made much simpler if the code is easily readable and well-docume
...
 
-2-2. We are writing C++11/14 with some restrictions.
-----------------------------------------------------
+2-2. We are writing C++14
+-------------------------
 
-The official policy on the use of C++11 features is at :ref:`Policy on use of C++11/14 language features <style-guide-cpp-cpp-11-14>`.
+The C++11 standard and the C++14 improvements to it bring a number of useful language features that make the resulting code more expressive, easier to read, and safer.
+
+We follow the official: International Standard ISO/IEC 14882:2014(E) – Programming Language C++, without any compiler specific extensions.
```
ktlim: I think we should say that the minimum compiler level is gcc 6.3.1 (and clang x.x.x?) and that if there are divergences between what the standard says should be supported and what the compilers support, the intersection set of portable features is what we allow.

coding/cpp_style_guide.rst
```diff
@@ -84,10 +84,17 @@ These tasks are made much simpler if the code is easily readable and well-docume
...
+
+.. seealso::
+
+   - :ref:`pipelines:source-install-redhat-legacy` from the `LSST Science Pipelines <https://pipelines.lsst.io>`__ documentation.
+   - :doc:`/services/lsst-dev` provides :ref:`instructions for using devtoolset-6 <lsst-dev-tools>` to obtain a more modern GCC on LSST cluster machines.
```
ktlim: Are there similar instructions for the minimum XCode level for macOS machines?

pschella: You need at least Xcode 6 on OS X version 10.9.4 (e.g clang 3.4). But I don't think there are instructions.

coding/cpp_style_guide.rst
pschella: Done

PR #176, opened 2017-10-30
DM-10343: Updated shared stacks.

jdswinbank:


services/lsst-dev.rst
```diff
@@ -116,39 +116,41 @@ Select Appropriate Developer Tools
...
+.. warning::
 
-.. prompt:: bash
-
-   scl enable devtoolset-3 git19 bash
+   Code compiled by different versions of GCC may not be compatible: it is generally better to stick to a particular toolchain for a given project.
+   In particular, if you are using a :ref:`shared stack <lsst-dev-loadlsst>` you *must* use the matching toolchain.
 
-You may wish to automatically enable ``devtoolset-3`` every time you log in to ``lsst-dev01`` by adding it to your shell initialization files.
+You may wish to automatically a particular Software Colllection every time you log in to ``lsst-dev01`` by adding it to your shell initialization files.
```
parejkoj: Per my comment on slack, we should be clear that this command *should not* go into `.bashrc`.

services/lsst-dev.rst
```diff
@@ -116,39 +116,41 @@ Select Appropriate Developer Tools
...
-The Developer Toolset includes version 1.9.3 of the `Git`_ version control system.
-If you prefer the (slightly) more recent version 1.9.4, you may also wish to enable the ``git19`` package.
-This may be done at the same time as enabling ``devtoolset-3``.
+.. warning::
 
-.. prompt:: bash
-
-   scl enable devtoolset-3 git19 bash
+   Code compiled by different versions of GCC may not be compatible: it is generally better to stick to a particular toolchain for a given project.
+   In particular, if you are using a :ref:`shared stack <lsst-dev-loadlsst>` you *must* use the matching toolchain.
```
parejkoj: Since all our stacks currently use the same toolchain, and this guide is specifically for DM developers, I wonder if we shouldn't be even more explicit here about what people should do:

```If you are using a shard stack<link>, put the following into your .profile: ```

or something like that. Plus, then we can be sure people get the correct `if hostname == blah` or
```bash
if [[ $- != *i* ]] ; then
    # Shell is non-interactive.
    return
fi
```
stuff so as to not get errors when running batch jobs (because `scl` is not available on the cluster).

jdswinbank: I'll try to rework the wording to address this concern to some extent. I'm worried that it's possible in general, though. We explicitly support multiple shells (we provide four different `loadLSST` scripts), and people have their environments configured in different ways: trying to explicitly document what to do in every possible circumstance is hard (and honestly, I've no idea what the correct incantations are in ksh, nor do I think it's a good use of time to find out).

Anyway, I'll see if I can come up with some sort of wording which is a halfway house.

parejkoj: I guess I'd say "provide a good solution for bash" (almost certainly the most common shell) to prevent most of us from having to come up with our own. There were some possible solutions on slack#dm-infrastructure to choose from, but most DM developers are likely not shell wizards (see my forkbombing), so a standard shell setup would be widely used.

parejkoj: Oh, and my comment about this also extend to getting a modern version of `git` setup as well

services/lsst-dev.rst
```diff
@@ -116,106 +116,139 @@ Select Appropriate Developer Tools
...
 
-You may wish to automatically enable ``devtoolset-3`` every time you log in to ``lsst-dev01`` by adding it to your shell initialization files.
-For example, try adding the following to :file:`~/.profile`:
+You may wish to automatically enable a particular software collection every time you log in to ``lsst-dev01`` and other LSST systems.
+Take care if you do this: it's easy to accidentally to either start recursively spawning shells and run out of resources or lock yourself out of machines which don't have the particular collection you're interested in installed.
+If you are using `Bash`_ — the default shell on ``lsst-dev01`` — try placing the following at the end of :file:`~/.bash_profile` and customising the list of ``desired_scls``.
```
jdswinbank: @parejkoj @mfisherlevine I think the below provides the sort of instructions you are looking for. Doing this in a way which doesn't break if `scl` is missing, if `devtoolset-n` is missing, if `devtoolset-n` is already activated, which enables the user to activate both `git19` and `devtoolset-n` at the same time, etc etc gets complicated, and I'm no shell wizard — could you let me know if this does the trick?

parejkoj: That looks pretty good, but putting it at the end of my `.bash_profile` (which loads `.bashrc` which has `source /software/lsstsw/stack/loadLSST.bash` at the end) does result in the following redundancy on login:

```
Note: use devtoolset-6 in conjunction with this stack.
Enabling git19 devtoolset-6.
Note: use devtoolset-6 in conjunction with this stack.
```

jdswinbank: Makes sense: your `.bash_profile` is sourcing `.bashrc` which sources `loadLSST.sh`, then `exec`ing another copy of `bash`, which once again sources `.bashrc` and hence `loadLSST.sh`. Probably fine to just add the `loadLSST.bash` line to your `.bash_profile` before the `exec`, I think.

PR #177, opened 2017-11-13
First attempt at writing down supported platforms

timj:
It turns out that we don't write down anywhere what operating systems we are really supporting. Here are some words for people to argue over.

### DIFF
--- a/build-ci/platforms.rst
+++ b/build-ci/platforms.rst
@@ -1,0 +1,27 @@
+###################
+Supported Platforms
+###################
+
+Introduction
+============
+
+The LSST Data Management software is required to be portable across many compute platforms ranging from high-performance computing centers to laptops (DMS-REQ-0308).
+The software will likely work on any Unix-like systems but we limit our testing and support to a fixed set of platforms.
+
+Patches may be accepted from developers who test on other operating systems.
+
+Platforms
+=========
+
+We have a baseline platform which is what we are using for integration testing and deployments.
+For each product, we also have a number of other platforms which are regularly tested by our continuous integration system to enhance portability.
+
+The baseline system is currently CentOS 7 on Intel x86_64 with gcc 6.3.1 from devtoolset-6 and Python 3.6.
+
+We regularly-test the following platforms in addition to the baseline platform:
+
+* CentOS 6 with devtoolset-6
+* macOS Sierra with Apple clang compilers
+* macOS High Sierra with Apple clang compilers
+
+See https://ci.lsst.codes for the current list of regularly-tested platforms.
--- a/index.rst
+++ b/index.rst
@@ -74,7 +74,7 @@ The `README <https://github.com/lsst-dm/dm_dev_guide/blob/master/README.md>`__ w
    docs/cpp_docs.rst
    docs/jsdoc.rst
    docs/technotes.rst
-   
+
 
 .. toctree::
    :maxdepth: 1
@@ -94,6 +94,7 @@ The `README <https://github.com/lsst-dm/dm_dev_guide/blob/master/README.md>`__ w
    :caption: Build, Test, Release
    :name: part-build-ci
 
+   build-ci/platforms.rst
    build-ci/eups_tutorial.rst
    build-ci/ci_overview.rst
    build-ci/lsstsw.rst


jdswinbank:
> people to argue over

👍 

What does "support" mean?

Actually, I ask that in two senses:

- You say we "support" CentOS 6 and various versions of macOS, but code is only required to work on CentOS 7. What does it mean to support platforms which our code might not work on?

- Does "support" imply an offer from the project to provide support to folks who want to get our code working on one of the supported platforms? I'm sure the answer ought to be "no", but "support" is quite an ambiguous term.

build-ci/platforms.rst
```diff
@@ -0,0 +1,33 @@
...
+Supported Platforms
+###################
+
+Introduction
+============
+
+The LSST Data Management software is required to be portable across many compute platforms ranging from high-performance computing centres to laptops (DMS-REQ-0308).
+The software will likely work on any Unix-like operating system but we limit our testing and support to a fixed set of operating systems.
+
+Patches will be accepted by developers who test on other operating systems.
```
jdswinbank: Perhaps "may"? When we receive a bazillion line patch set updating our code to work on Xenix, I doubt we're going to accept it.

timj:
"Support". What do we really mean? It's not entirely clear if "support" is the same for all of the stack. @fritzm could declare that Qserv only has to run on CentOS7 and ignore macOS and CentOS6. "Support" means we'll try to make sure our software works on all the supported platforms but if we are going down a rabbit hole trying to keep CentOS6 going we can drop it (dropping CentOS7 would be a lot harder to justify). Similarly, when SIP turned up on macOS as a project we were okay with not spending any effort on macOS for a few months.

So maybe "support" means "we will endeavor to ensure that our software works on all supported platforms" but we can decide to drop any non-CentOS7 platforms if they are becoming too much of a burden and we can drop specific subsets of "the stack" from some platforms.

"Supported" for actual releases is a definite statement that that suite of software that was part of that release definitely did work on those operating systems.

timj:
I've had another crack at the text.

build-ci/platforms.rst
```diff
@@ -0,0 +1,46 @@
...
+Supported Operating Systems
+===========================
+
+The following operating systems are supported and tested by Jenkins.
+
+CentOS 7
+--------
+
+The baseline operating system is CentOS7 and all code *must* work on CentOS 7.
+We currently use the default CentOS 7 compiler.
```
ktlim: For development, integration testing, and deployment, I believe we currently use devtoolset-6.

build-ci/platforms.rst
```diff
@@ -0,0 +1,46 @@
...
+We aim to support the most recent versions of CentOS7 within 6 months of release, although we can not guarantee that all NCSA computers and Jenkins will be upgraded promptly.
```
ktlim: Is it really useful to state this, given the caveat?

build-ci/platforms.rst
```diff
@@ -0,0 +1,46 @@
...
+
+CentOS 6
+--------
+
+We currently support CentOS 6 with the devtoolset-3 compiler set.
+
+macOS
+-----
+
+We support the two most recent releases of the Apple macOS operating system using the Apple clang compilers.
```
ktlim: We don't test against the two most recent, do we?  And which XCodes do we actually test with?

build-ci/platforms.rst
```diff
@@ -0,0 +1,46 @@
...
+Patches may be accepted from developers who test on other operating systems.
+
+Definitions
+===========
+
+We use the following definitions when defining which operating systems are supported.
+
+* A *supported* operating system is one which is regularly tested by our continuous integration system and which we will ensure all tests pass.
+  The list of supported operating systems may change over time, an in particular, if we encounter difficulties where keeping an operating system running is taking excessive developer and CI resources, we reserve the right to drop it from the list.
+  Each supported operating system will be associated with a specific compiler that we use for testing.
```
ktlim: We should be testing using multiple compilers on CentOS, including clang and icc.

build-ci/platforms.rst
```diff
@@ -0,0 +1,46 @@
...
+
+The LSST Data Management software is required to be portable across many compute platforms ranging from high-performance computing centres to laptops (DMS-REQ-0308).
+The software will likely work on any Unix-like operating system but we limit our testing and support to a fixed set of operating systems.
+
+Patches may be accepted from developers who test on other operating systems.
+
+Definitions
+===========
+
+We use the following definitions when defining which operating systems are supported.
```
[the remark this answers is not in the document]
timj: I thought technically we were still not using devtoolset-6 because jenkins is still not up to date.

build-ci/platforms.rst
timj: At one point @billglick mentioned that it would be good to state our actual upgrade policy but I'm happy to drop it from here and find somewhere else to say it.

build-ci/platforms.rst
timj: Apple doesn't give you a choice -- you get the XCode you get from the App Store based on the OS version. We test Sierra and many people run High Sierra. That seems to be good enough to say we support both.

build-ci/platforms.rst
timj: yes, but we aren't at the moment. Maybe if we dropped centos6 we could add a clang job instead.

build-ci/platforms.rst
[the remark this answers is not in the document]
timj: Scratch that. You are saying baseline means specifically NCSA. Jenkins is testing "supported" platforms but not at the moment the baseline platform.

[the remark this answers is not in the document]
timj: Are you suggesting that your paragraph replace most of the document?

timj:
I've rejigged it based on @ktlim's feedback.

PR #184, opened 2017-11-28
Document `gen3-middlware` label.

jdswinbank:


jdswinbank:
Trivial change merged without review.

PR #185, opened 2017-12-01
Update lsst-dev.rst

billglick:
Added rh-git29 as SCL package


jdswinbank:
Been using this for a couple of weeks with no problems; I think we can go ahead and merge.

PR #188, opened 2017-12-12
drp: update devtoolset-6 activation

PaulPrice:
"module load" is better than "scl enable".

### DIFF
--- a/teams/drp.rst
+++ b/teams/drp.rst
@@ -55,7 +55,7 @@ You should therefore enable ``devtoolset-6`` before proceeding:
 
 .. prompt:: bash
 
-  scl enable devtoolset-6 bash
+  module load rh/devtoolset/6
 
 For more information on devtoolset usage, refer to :ref:`the main developer guide <lsst-dev-tools>`.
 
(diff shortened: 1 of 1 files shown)


jdswinbank:
Why? (I'm not denying it being better, I just genuinely don't know the reasoning.)

PaulPrice:
It doesn't throw you into a subshell.

jhoblitt:
The preferred scl incantation is  `source scl_source enable devoolset-6` and that should generally work on centos-7.  I have found that `scl_source` from the version of `scl-utils` on centos-6 can fail when used non-interactively.  `source /opt/rh/devtoolset-6/enable` seems to always be reliable but I'm unsure if that would cause problems when using multiple scls simultaneously.

jdswinbank:
```
$ head -5 `which scl_source` 
#!/bin/bash

_scl_source_help="Usage: source scl_source <action> [<collection> ...]

Don't use this script outside of SCL scriptlets!
```
I don't think we want to adopt that as the "preferred" incantation.

jhoblitt:
That is the official solution from the scl authors:

https://access.redhat.com/solutions/527703

PaulPrice:
On the Princeton systems referred to, the admins have provided this as a `module`, so I think that should be preferred (on these systems only) because it will be maintained by the admins, is sure to include anything these systems need, and it avoids the questions about whether `scl_source` should be used.

PR #189, opened 2017-12-15
DM-12951: Improve Numpydoc documentation

jonathansick:
This PR includes many improvements to the Numpydoc documenation for developers. See the draft online at https://developer.lsst.io/v/DM-12951/docs/py_docs.html

- Clarified formatting of Returns section (https://jira.lsstcorp.org/browse/DM-12926)
- Clarified formatting of Parameters section (https://jira.lsstcorp.org/browse/DM-12936)
- More accurate recommentations for attribute documentation based on real-world experience.
- Better and more accurate examples.
- Lots of copy edits.

### DIFF
--- a/docs/py_docs.rst
+++ b/docs/py_docs.rst
@@ -865,7 +865,13 @@ Module docstrings contain the following sections:
 3. :ref:`Extended Summary <py-docstring-extended-summary>` (optional)
 4. :ref:`See Also <py-docstring-see-also>` (optional)
 
-.. TODO Provide an example
+.. note::
+
+   Module docstrings aren't featured heavily in the documentation we generate and publish with Sphinx.
+   Avoid putting important end-user documentation in module docstrings.
+   Instead, write introductory and overview documentation in the module's *user guide* (the :file:`doc/` directories of Stack packages).
+
+   Module docstrings can still be useful for developer-oriented notes, though.
 
 .. _py-docstring-class-structure:
 
--- a/docs/py_docs.rst
+++ b/docs/py_docs.rst
@@ -338,15 +338,13 @@ The sections and their relative order is:
 2. :ref:`Deprecation Warning <py-docstring-deprecation>` (if applicable)
 3. :ref:`Extended Summary <py-docstring-extended-summary>` (optional)
 4. :ref:`Parameters <py-docstring-parameters>` (if applicable; for classes, methods, and functions)
-5. :ref:`Methods <py-docstring-methods>` (if applicable; for classes)
-6. :ref:`Attributes <py-docstring-attributes>` (if applicable; for classes)
-7. :ref:`Returns <py-docstring-returns>` or :ref:`Yields <py-docstring-yields>` (if applicable; for functions, methods, and generators)
-8. :ref:`Other Parameters <py-docstring-other-parameters>` (if applicable; for classes, methods, and functions)
-9. :ref:`Raises <py-docstring-raises>` (if applicable)
-10. :ref:`See Also <py-docstring-see-also>` (optional)
-11. :ref:`Notes <py-docstring-notes>` (optional)
-12. :ref:`References <py-docstring-references>` (optional)
-13. :ref:`Examples <py-docstring-examples>` (optional)
+5. :ref:`Returns <py-docstring-returns>` or :ref:`Yields <py-docstring-yields>` (if applicable; for functions, methods, and generators)
+6. :ref:`Other Parameters <py-docstring-other-parameters>` (if applicable; for classes, methods, and functions)
+7. :ref:`Raises <py-docstring-raises>` (if applicable)
+8. :ref:`See Also <py-docstring-see-also>` (optional)
+9. :ref:`Notes <py-docstring-notes>` (optional)
+10. :ref:`References <py-docstring-references>` (optional)
+11. :ref:`Examples <py-docstring-examples>` (optional)
 
 For summaries of how these docstring sections are composed in specific contexts, see:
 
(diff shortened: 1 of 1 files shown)


docs/py_docs.rst
```diff
@@ -556,111 +745,88 @@ Raises
...
 
-'See Also' is an optional section used to refer to related code.
-This section can be very useful, but should be used judiciously.
-The goal is to direct users to other functions they may not be aware of, or have easy means of discovering (by looking at the module docstring, for example).
-Routines whose docstrings further explain parameters used by this function are good candidates.
+The description text is indented by four spaces from the docstring's left justification.
+Like the description fields for :ref:`Parameters <py-docstring-parameters>` and :ref:`Returns <py-docstring-returns>`, the description can consist of multiple paragraphs and lists.
 
-As an example, for a function such as ``numpy.cos``, we would have
+Stylistically, write the first sentence of each description in the form:
```
kfindeisen: Good idea to standardize this. :+1:

docs/py_docs.rst
```diff
@@ -730,57 +901,61 @@ Module docstrings contain the following sections:
 3. :ref:`Extended Summary <py-docstring-extended-summary>` (optional)
 4. :ref:`See Also <py-docstring-see-also>` (optional)
 
-.. TODO Provide an example
+.. note::
+
+   Module docstrings aren't featured heavily in the documentation we generate and publish with Sphinx.
+   Avoid putting important end-user documentation in module docstrings.
+   Instead, write introductory and overview documentation in the module's *user guide* (the :file:`doc/` directories of Stack packages).
```
kfindeisen: Link to "Documenting Stack Packages" (or whatever will replace it)?

docs/py_docs.rst
```diff
@@ -226,7 +223,31 @@ Class/method/function docstrings must be placed directly below the declaration,
...
+The class docstring takes the place of a docstring for the ``__init__`` method.
+``__init__`` methods don't have docstrings.
+
+.. _py-docstring-attributes:
+
+Module Constants and Class Attributes
+-------------------------------------
+
+Docstrings for module-level variables and class attributes appear directly below their first declaration.
+For example:
```
kfindeisen: You might want to link to or mention the policy on attributes defined by `__init__`, since that confused me when I was reading this.

docs/py_docs.rst
```diff
@@ -838,35 +1042,73 @@ At minimum, constants/properties/attributes should have a summary line, but can
...
+
+       @quantity.setter
+       def quantity(self, q):
+           # ...
+
+Note:
+
+- Do not use the :ref:`Returns section <py-docstring-returns>` in the property's docstring.
+  Instead, include type information in the summary, as is done for :ref:`class attributes <py-docstring-attribute-constants-structure>`.
+- Along with type information, include whether the property is "immutable" or "mutable" (has a setter method).
```
kfindeisen: Can the user distinguish between attributes and properties in the generated documentation? Otherwise, it seems odd to make a property say "mutable" while not saying anything for an equivalent attribute.

Also, maybe Python usage of the term is different, but I would say that a property is unmodifiable or read-only, not immutable -- in general its value *can* change, just not by assignment through the property name.

docs/py_docs.rst
```diff
@@ -838,35 +1042,73 @@ At minimum, constants/properties/attributes should have a summary line, but can
...
-   """Name of the project (str)"""
+   """Name of the project (`str`)."""
 
-Note that class attributes can alternatively be documented in an :ref:`Attributes <py-docstring-attributes>` section of the class's docstring.
-This is particularly useful when the attribute is not set in the class scope, but rather in a method such as ``__init__``.
+Note:
 
-.. code-block:: py
+- The docstring appears directly below the constant or class attribute.
+- The type is included in parentheses at the end of the summary line.
```
kfindeisen: Consider an example with more than just a summary line (I assume the type would still be at the end of the first line, not the end of the docstring).

docs/py_docs.rst
```diff
@@ -250,70 +271,134 @@ For example
...
+------------
+
+Hard-wrap text in docstrings to match the line length allowed by the coding standard (see :ref:`style-guide-py-line-length`).
+
+.. note::
 
-In docstrings, the top level header is marked up with a ``-``, the third level listed in our :ref:`ReStructuredText Style Guide <rst-sectioning>`.
-The header hierarchy is thus:
+   In the future we may be able use a shorter, more readable, line length for docstrings.
+   See :jira:`RFC-107`.
```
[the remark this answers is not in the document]
jonathansick: That would be the goal, and I do agree about 80 or so characters for docstrings. I think that by default folks will wrap at the 110 character column they've configured their editors to show. Maybe I'll change the wording to be more permissive towards those trying to be future compatible with the next-generation standard.

docs/py_docs.rst
```diff
@@ -368,78 +459,170 @@ Parameters
...
+       - ``x``: current parameter (`float`).
+       - ``extra_args``: additional arguments (`dict`).
 
 .. _py-docstring-optional:
 
 Optional Parameters
 ^^^^^^^^^^^^^^^^^^^
 
-For keyword arguments, add 'optional' to the type specification:
+For keyword arguments with useful defaults, add ``optional`` to the type specification:
```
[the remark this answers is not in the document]
jonathansick: It's a convention that comes from the Numpy project and I think it's a good idea.

docs/py_docs.rst
jonathansick: Good idea; I'll only have properties without setter say they are "read-only." That should fix the terminology and also add consistency with regular class attributes.

jonathansick:
@kfindeisen, thanks for the feedback.

I've also moved the "placement" documention into the sections about individual types of docstrings. I think it's more usable/less confusing that way.

docs/py_docs.rst
```diff
@@ -556,111 +649,88 @@ Raises
...
 .. code-block:: rst
 
-   `astropy.table.Tables` : Flexible table data structures
+   See also
+   --------
+   sin
+   tan
 
-Functions may be listed without descriptions; this is preferable if the functionality is clear from the function name:
+Numpydoc assumes that the contents of the 'See also' section are API names, so don't wrap each name with backticks, as we do when describing types in :ref:`Parameters <py-docstring-parameters>` and :ref:`Returns <py-docstring-returns>`)
```
ktlim: Need a period at the end of this sentence.

docs/py_docs.rst
```diff
@@ -556,111 +649,88 @@ Raises
...
 
-When referring to an entirely different module or package, use the full namespace.
+As an example, for a function such as ``numpy.cos``, we would have:
 
 .. code-block:: rst
 
-   `astropy.table.Tables` : Flexible table data structures
+   See also
+   --------
+   sin
```
ktlim: Descriptions got dropped here; is the implication that they should never be used in "See Also"?

docs/py_docs.rst
```diff
@@ -677,21 +747,23 @@ References cited in the :ref:`'Notes' <py-docstring-notes>` section may be liste
       and neural-network techniques," Computers & Geosciences, vol. 22,
       pp. 585-588, 1996.
 
-Note that Web pages should be referenced with regular inline links.
+Web pages should be referenced with regular inline links.
```
ktlim: In the line above the example, "as in the list as follows" should perhaps be simply "as follows".

docs/py_docs.rst
```diff
@@ -838,35 +1043,118 @@ At minimum, constants/properties/attributes should have a summary line, but can
...
+
+   class MyClass(object):
+       """[...]
+       """
+
+       x = None
+       """Description of x attribute.
+       """
+
+Examples of Constant and Class Attribute Docstring Examples
```
ktlim: Extra "Examples" here, and "Docstring" should be plural.

docs/py_docs.rst
jonathansick: The Numpydoc developers originally documented descriptions in the *See also* section, but in reality, those have never been parsed properly.

docs/py_docs.rst
```diff
@@ -39,7 +35,7 @@ Docstrings MUST be delimited by triple double quotes
 
 Docstrings **must** be delimited by triple double quotes: ``"""``.
 This allows docstrings to span multiple lines.
-You may use ``u"""`` for unicode but it is usually preferable to stick to ASCII.
+You may use ``u"""`` for unicode, but it's usually preferable to stick to ASCII.
```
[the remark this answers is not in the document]
jonathansick: We'll have to include this in the list of docs to update when v15 comes out and we fully drop Python 2.7 compatibility. I think you're right that unicode docstrings should work in Python 3.

PR #191, opened 2017-12-21
Renamed lsst-db to lsst-dev-db

billglick:


services/lsst-db.rst
```diff
@@ -1,13 +1,13 @@
 ########################
-Using the lsst-db Server
+Using the lsst-dev-db Server
```
jonathansick: You you update the under/overline to match the title length?

```rst
############################
Using the lsst-dev-db Server
############################
```

PR #194, opened 2018-01-09
Added ToC entry mentioning Python 3.6

gpdf:
This parallels the C++ style guide and makes the reference version much easier to spot for the casual reader.

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -25,6 +25,12 @@ All DM Python code MUST work with Python 3
 
 All the Python code written by LSST Data Management must be runnable using Python 3.
 Python 2 will cease to be supported before LSST is operational (:pep:`373`).
+
+.. _style-guide-py-version-py36:
+
+We are writing Python 3.6
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


timj:
We are only writing python 3.6 in some cases. The baseline can be 3.6 without people being allowed to use 3.6 features. Until we drop python2.7 there is ambiguity here.

gpdf:
Are you suggesting I add some wording to make that clearer?  I can do that.

timj:
I'm happy for you to give it another try 😄 . Your statement will be correct in a few months. Otherwise you have to say something like:
> When not writing code that has to be compatible with Python 2.7, we are writing Python 3.6.

ktlim:
I think this whole section needs a rewrite/trim now that we're past v15, no?

jdswinbank:
The current Dev Guide makes this information [reasonably clear|https://github.com/lsst-dm/dm_dev_guide/blame/536e7395413e78f679cf8b9eaa2dda55e06b9812/python/style.rst#L28], and this POR has been hanging around for the better part of a year with no updates. I'm closing it down. Please reopen if you disagree with that decision.

PR #197, opened 2018-01-28
DM-7994: Add recommendations on compiler-generated members.

TallJimbo:
Implements RFC-209.  Text has been modified slightly from original RFC proposal to include destructors and incorporate discussion on documentation.

### DIFF
--- a/coding/cpp_style_guide.rst
+++ b/coding/cpp_style_guide.rst
@@ -1809,6 +1809,53 @@ Braced initialization may also be used in any context where the C++98 initializa
 While all of the above examples show objects being initialized in regular code, the same guidelines apply to the initialization of data members in an initialization list (though the assignment-like syntax can not be used in that context).
 
 
+.. _style-guide-cpp-5-27b:
+
+5-27b. Classes SHOULD explicitly declare or delete compiler-generated members
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+All new C++ classes SHOULD explicitly default or delete all constructors and assignment operators that may be synthesized by the compiler.
+Similarly, destructors should be explicitly defined using the ``default`` keyword as well when no custom implementation is needed.
+In practice, that means all classes should explicitly define the following five members using either ``default``, ``delete``, or a custom implementation (shown here defaulted):
+
+.. code-block:: cpp
+
+    class Foo {
+    public:
+        ...
+        Foo(Foo const &) = default;
+        Foo(Foo &&) = default;
+        Foo & operator=(Foo const &) = default;
+        Foo & operator=(Foo &&) = default;
+        ~Foo() = default;
+        ...
+    };
+
+Note that when using ``default``, these can be declared in the header and defined in a source file:
+
+.. code-block:: cpp
+
+    // in .h
+    class Foo {
+    public:
+        ...
+        Foo(Foo const &);
+        ...
+    };
+
+    // in .cc:
+    Foo::Foo(Foo const &) = default;
+
+The latter is slightly more verbose, but it protects downstream code from sensitivity to changes in the implementation of these members.
+
+Defining these members forces us to consider their behavior and judge whether the default behavior is in fact desired when implementing a new class.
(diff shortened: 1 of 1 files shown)


coding/cpp_style_guide.rst
```diff
@@ -1242,6 +1242,36 @@ A class or struct definition must explicitly declare the privacy qualifier of it
...
 This Rule will reduce unnecessary compiler warnings.
 
+.. _style-guide-cpp-5-2c:
+
+5-2c. Classes SHOULD explicitly declare or delete compiler-generated members
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+All new C++ classes SHOULD explicitly default or delete all constructors and assignment operators that may be synthesized by the compiler.
+Similarly, destructors should be explicitly defined using the ``default`` keyword as well when no custom implementation is needed.
+In practice, that means all classes should explicitly define the follow six members using either ``default``, ``delete``, or a custom implementation (shown here defaulted):
```
timj: "follow" -> "following"

TallJimbo: I also see that it should be "five", not "six" (at one point I had a defaulted default constructor, too, but I remove that as I didn't want to imply that all classes should have a default constructor).

coding/cpp_style_guide.rst
```diff
@@ -1242,6 +1242,36 @@ A class or struct definition must explicitly declare the privacy qualifier of it
 Although C++ provides the above definitions as defaults, some compilers generate warnings if explicit privacy qualifiers are not specified.
 This Rule will reduce unnecessary compiler warnings.
 
+.. _style-guide-cpp-5-2c:
```
ktlim: I think this fits better as 5-27b, even though that's further down.

coding/cpp_style_guide.rst
```diff
@@ -1242,6 +1242,36 @@ A class or struct definition must explicitly declare the privacy qualifier of it
...
+        Foo & operator=(Foo const &) = default;
+        Foo & operator=(Foo &&) = default;
+        ~Foo() = default;
+        ...
+    };
+
+Defining these members forces us to consider their behavior and judge whether the default behavior is in fact desired when implementing a new class.
+It also makes compilation errors appear sooner (rather than delaying them until a synthesized member is first needed).
+
+When these members cannot be defaulted or deleted, a comment explaining why SHOULD accompany the implementation.
```
ktlim: Should a note be added for `delete` also?  That would also help to interrupt the reader who might otherwise not notice the distinction between medium-length words that start with "de" and will be more boilerplate-ish given this guideline.  In many cases it seems to me that custom implementations are straightforward and don't necessarily need explanation, but I'm OK with asking for one anyway.

TallJimbo: Done.  I believe @r-owen liked the idea of asking for comments when using `delete`, too, and while I do feel like there are times when it's perfectly clear without a comment, I certainly agree that there are times when it isn't.

coding/cpp_style_guide.rst
```diff
@@ -1242,6 +1242,36 @@ A class or struct definition must explicitly declare the privacy qualifier of it
...
+    class Foo {
+    public:
+        ...
+        Foo(Foo const &) = default;
+        Foo(Foo &&) = default;
+        Foo & operator=(Foo const &) = default;
+        Foo & operator=(Foo &&) = default;
+        ~Foo() = default;
+        ...
+    };
```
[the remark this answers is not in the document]
TallJimbo: How about I add a comment in the example code noting that they can also go in source files?  I'm not sure I want to advocate that they *should* go in the source files, but I certainly don't want to inadvertently advocate against it.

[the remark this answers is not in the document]
TallJimbo: I actually went ahead and added an extra example - it was just hard to make the comment version sufficiently clear.

coding/cpp_style_guide.rst
```diff
@@ -1809,6 +1809,53 @@ Braced initialization may also be used in any context where the C++98 initializa
...
+
+Note that when using ``default``, these can be also declared in the header and defined in a source file:
```
ktlim: I think "also" can be omitted here, or at a minimum moved to before "be".

PR #198, opened 2018-01-29
DM-13399: Use cls for first argument to Python metaclass methods

pschella:


coding/python_style_guide.rst
```diff
@@ -553,6 +553,21 @@ When tempted to use ``l``, use ``L`` instead.
...
 
+.. _style-guide-py-naming-metaclasses:
+
+Always use ``cls`` for the first argument to metaclass instance methods
+-----------------------------------------------------------------------
+
+For regular classes ``self`` is used, but for class methods and hence also for metaclass instance
+methods, ``cls`` should be used instead.
+
+Also add ``# noqa N805`` to the relevant line in order to support older versions of flake8.
```
ktlim: "older" begs the question of "than what?"...

If a fixed flake8 is about to hit, should we wait, update the stack, and then not need this line at all?

timj: We need PyCQA/pep8-naming#47 to end up in a release of pep8-naming so it might not be a long wait.

timj:
In theory we could pretty easily update our pep8-naming to be v0.5.0 with the upstream patches applied to do this (they are the only extra commits on master). So we could do this without `#noqa` quite quickly if we were motivated.

pschella:
I removed the `noqa` line. We can merge this when upstream is ready.

ktlim:
Any idea when a new version of pep8-naming might be released?  Given their history, it looks like it could be a year or more.  We could go with the 0.5.0 + patch, but I'm a bit worried that in-editor linters might not pick up that version.

timj:
There is a thread in PyCQA/pep8-naming#47 but there has been no feedback from the developer.

PR #200, opened 2018-02-07
Update pointer to sdss dataset

SimonKrughoff:


SimonKrughoff:
I've filed DM-13508 for making a catalog.

PR #201, opened 2018-02-14
DM-13525: Improve numpydoc examples

jonathansick:
https://developer.lsst.io/v/DM-13525/docs/py_docs.html

- Clarify that class attributes and parameters are documented separately.
- Example of returning an `lsst.pipe.base.Struct`.
- Full example module (based on http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy)
- Easier to use table of contents

### DIFF
--- a/docs/py_docs.rst
+++ b/docs/py_docs.rst
@@ -8,16 +8,36 @@ Documenting Python APIs with Docstrings
 
 We use Python docstrings to create reference documentation for our Python APIs.
 Docstrings are read by developers, interactive Python users, and readers of our online documentation.
-This page describes how to write these docstrings in Numpydoc, DM's standard format:
+This page describes how to write these docstrings in Numpydoc, DM's standard format.
+
+**Format reference:**
 
 - :ref:`py-docstring-basics`.
 - :ref:`py-docstring-rst`.
 - :ref:`py-docstring-sections`.
+
+  1. :ref:`Short Summary <py-docstring-short-summary>`
+  2. :ref:`Deprecation Warning <py-docstring-deprecation>`
+  3. :ref:`Extended Summary <py-docstring-extended-summary>`
+  4. :ref:`Parameters <py-docstring-parameters>`
+  5. :ref:`Returns <py-docstring-returns>` or :ref:`Yields <py-docstring-yields>`
+  6. :ref:`Other Parameters <py-docstring-other-parameters>`
+  7. :ref:`Raises <py-docstring-raises>`
+  8. :ref:`See Also <py-docstring-see-also>`
+  9. :ref:`Notes <py-docstring-notes>`
+  10. :ref:`References <py-docstring-references>`
+  11. :ref:`Examples <py-docstring-examples>`
+
+**How to format different APIs:**
+
 - :ref:`py-docstring-module-structure`.
 - :ref:`py-docstring-class-structure`.
 - :ref:`py-docstring-method-function-structure`.
 - :ref:`py-docstring-attribute-constants-structure`.
 - :ref:`py-docstring-property-structure`.
+
+**Learn by example:**
+
 - :ref:`py-docstring-example-module`.
 
 Treat the guidelines on this page as an extension of the :doc:`../coding/python_style_guide`.
--- a/docs/py_docs.rst
+++ b/docs/py_docs.rst
@@ -1216,10 +1216,6 @@ The best way to document these public attributes is by declaring the attribute a
 
 Notice that the :ref:`parameters <py-docstring-parameters>` to the ``__init__`` method are documented separately from the class attributes (highlighted).
 
-.. note::
-
-   Private attributes (prefixed by underscores: ``self._myAttribute``) do not need to be documented with docstrings.
-
 .. _py-docstring-property-structure:
 
 Documenting Class Properties
--- a/docs/py_docs.rst
+++ b/docs/py_docs.rst
@@ -1070,12 +1070,18 @@ For example:
 Examples of Constant and Class Attribute Docstrings
 ---------------------------------------------------
 
+Minimal constant or attribute example
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 Include the attribute or constant's type in parentheses at the end of the summary line:
 
 .. code-block:: py
 
    NAME = 'LSST'
    """Name of the project (`str`)."""
+
+Multi-section docstrings
+^^^^^^^^^^^^^^^^^^^^^^^^
 
 Multi-section docstrings keep the type information in the summary line.
 For example:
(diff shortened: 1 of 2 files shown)


jonathansick:
@PaulPrice Does this Struct return type example help? https://developer.lsst.io/v/DM-13525/docs/py_docs.html#struct-return-types

docs/py_docs.rst
```diff
@@ -909,7 +984,7 @@ Here's an example of a class:
        ------
        ValueError
            Raised when input angles are outside range.
-
+
```
timj: trailing whitespace.

docs/py_docs.rst
```diff
@@ -1097,28 +1179,43 @@ For example:
...
 The best way to document these public attributes is by declaring the attribute at the class level and including a docstring with that declaration:
 
 .. code-block:: python
+   :emphasize-lines: 12-18
 
    class Metric(object):
        """Verification metric.
 
-       [...]
+       Parameters
```
timj: might be good to have a parameter here in the example that is not also an attribute.

docs/py_docs.rst
```diff
@@ -1097,28 +1179,43 @@ For example:
...
+       def __init__(self, name, unit):
            self.name = name
            self.unit = unit
 
-Private attributes (prefixed by underscores: ``self._myAttribute``) do not need to be documented with docstrings.
+Notice that the :ref:`parameters <py-docstring-parameters>` to the ``__init__`` method are documented separately from the class attributes (highlighted).
+
+.. note::
+
+   Private attributes (prefixed by underscores: ``self._myAttribute``) do not need to be documented with docstrings.
```
timj: How do developers of the class (as opposed to users of the class) know what they are for then? If you document it with a docstring I thought it was hidden by default but you could use a flag to turn it on. I think it is good practice to document private methods and private attributes and if we don't give guidance here we'll have a mix of some with docstrings, some with commented text and some completely undocumented.

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,297 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""Example Python module with Numpydoc-formatted docstrings.
+
+This module demonstrates documentation written according to LSST DM's guidelines for `Documenting Python
```
timj: I think these lines are too long for docstrings. I know we are still struggling with @brianv0's patch to flake8 but I really think our example docstrings should be under 80 characters if we are to have any ability to get people to use shorter lines.

PaulPrice:
That's good, but can I suggest that the following would be more in keeping with our general usage pattern:

```
def processSomething(self):
    """Process something

    Returns
    -------
    result : `lsst.pipe.base.Struct`
        Result struct with components:
 
        - ``thing`` : The thing at the place with the guy (`str`)
        - ``place`` : The place for the thing with the guy (`int`)
        - ``guy`` : The guy at the place with the thing (`str`)
    """
```

It's such a shame this pattern is so awkward (empty line, double backticks, type not required) as we will use it a lot. Could we say that if we just return a single `Struct`, we can document that as multiple return values?

```
def processSomething(self):
    """Process something

    Returns a struct of results.

    Returns
    -------
    thing : `str`
        The thing at the place with the guy.
    place : `int`
        The place for the thing with the guy.
    guy : `str`
        The guy at the place with the thing.
    """
```

It's a little misleading (we're documenting multiple return values but there's only one), but it's **so** much nicer to write.

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Returns
+    -------
+    success : `bool`
+        `True` if successful, `False` otherwise.
+
+        The return type is not optional. The ``Returns`` section may span multiple lines and paragraphs.
+        Following lines should be indented to match the first line of the description.
+
+        The ``Returns`` section supports any reStructuredText formatting, including literal blocks::
```
timj: Is it the `::` that indicates the next indented section is a literal block?

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,297 @@
...
+    def __init__(self, param1, param2, param3):
+        self.attr1 = param1
+        self.attr2 = param2
+        self.attr3 = param3
+
+        self.attr4 = ['attr4']
+
+    @property
+    def readonlyProperty(self):
+        """Properties are documented in their getter method (`str`, read-only).
```
timj: This is crying out for function annotations:
```python
@property
def readonlyProperty(self) -> str:
   ...
```

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,297 @@
...
+        If the setter method contains notable behavior, it should be mentioned here as well.
+        """
+        return self.attr4
+
+    @readwriteProperty.setter
+    def readwriteProperty(self, value):
+        self.attr4 = value
+
+    def exampleMethod(self, param1, param2):
+        """Test that a sitaution is true.
```
timj: typo: situation

jonathansick:
@PaulPrice If you refresh https://developer.lsst.io/v/DM-13525/docs/py_docs.html#struct-return-types does that get closer to what you'd like? Using the term "components" seemed to be the main difference. Unfortunately, we can't use your suggested format until we decide to fork the Numpydoc parser.

PaulPrice:
Good enough.

Does the numpydoc parser actually check that the number of objects being returned matches what's in the doc? Otherwise, what's wrong with adopting my suggested form?

docs/snippets/numpydocExample.py
jonathansick: I guess it doesn't hurt to have the examples shorter. We can just remain silent on whether the regular-length docstring lines are required or not.

timj: Yes, we really don't want to go back and fix these examples later when flake8 finally does what we want.

jonathansick:
> Does the numpydoc parser actually check that the number of objects being returned matches what's in the doc? Otherwise, what's wrong with adopting my suggested form?

The inconsistency will lead to confusion unless we make returning a Struct a special case. An acceptable solution would be forking Numpydoc to add a new section that replaces ``Returns`` and ``Yields`` with:

```
Result struct
-------------
componentA : `int`
    Description of ``component A``.
componentB : `int`
    Description of ``component B``.
```

But going down this route is off-topic for the present PR. Thanks for your feedback.

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,323 @@
...
+    specification. See `Documenting Methods and Functions`_ for more
+    information about function docstrings.
+
+    Parameters
+    ----------
+    param1 : `int`
+        The first parameter. Note how the type is marked up with backticks.
+        This marks ``int`` as an API object so that Sphinx will attempt to
+        link to its reference documentation. You can do this for custom types
+        as well. You'll see an example in the *Returns* section.
```
ktlim: I don't see a custom type in *Returns*.

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,323 @@
...
+    param2 : `str`, optional
+        Optional arguments (those with defaults) always include the word
+        ``optional`` after the type info. See the `Parameters`_ section
+        documentation for details.
+    *args
+        Variable length argument list.
+    **kwargs
```
ktlim: If we're trying to give a complete example, shouldn't there be documentation on a keyword argument here?

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,323 @@
...
+
+    attr3 = None
+    """Description of ``attr3`` (`int`).
+    """
+
+    attr4 = None
+    """Description of ``attr4`` (`list` of `str`).
+    """
+
+    def __init__(self, param1, param2, param3):
```
ktlim: Doesn't look like `param3` is optional here.

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,323 @@
...
+            `True` if successful, `False` otherwise.
+
+        Notes
+        -----
+        Do not include the ``self`` parameter in the ``Parameters`` section.
+        """
+        return True
+
+    def __special__(self):
+        """At the moment, special members with docstrings are not published in
```
ktlim: I hope this is going to be fixed; it seems like a substantial limitation.  Nevertheless, we should make clear that such members _must_ be documented anyway.

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,323 @@
...
+        However, you should still provide docstrings for private members to
+        document code for internal developers.
+        """
+        pass
+
+
+class ExampleError(Exception):
+    """Example exception.
+
+    Exceptions are documented in the same manner other classes.
```
ktlim: "manner _as_ other"

docs/snippets/numpydocExample.py
```diff
@@ -0,0 +1,323 @@
...
+        Human readable string describing the exception.
+    code : `int`, optional
+        Numeric error code.
+
+    Notes
+    -----
+    Do not include the ``self`` parameter in the ``Parameters`` section.
+    """
+
+    msg = None
```
ktlim: It's unfortunate that documenting constructed state is so repetitive.

PR #202, opened 2018-02-15
DM-5383: Update copyright instructions

timj:
Follows the discussion in RFC-45. I've also tried to factor out the license text that is present in each source in a single location referenced from C++ and Python guides.

### DIFF
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -235,11 +235,8 @@ Each Python file MUST contain the standard license preamble
 A copyright and license block using :ref:`the standard text <pkg-doc-code-preamble>` MUST be included at the top of each file.
 This can be written as a Python comment.
 
-   .. code-block:: python
-
-      #
-      # <template text>
-      #
+.. literalinclude:: ../docs/snippets/license_preamble.txt
+   :language: text
 
 .. _style-guide-py-line-length:
 
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -227,6 +227,20 @@ Style changes must be encapsulated in a distinct commit (see :ref:`git-commit-or
 
    :doc:`../docs/py_docs` provides guidelines for the :ref:`layout of docstrings <py-docstring-basics>`.
 
+.. _style-guide-license:
+
+Each Python file MUST contain the standard license preamble
+-----------------------------------------------------------
+
+A copyright and license block using :ref:`the standard text <pkg-doc-code-preamble>` MUST be included at the top of each file.
+This can be written as a Python comment.
+
+   .. code-block:: python
+
+      #
+      # <template text>
+      #
+
 .. _style-guide-py-line-length:
 
 Line Length MUST be less than or equal to 110 columns
--- a/coding/python_style_guide.rst
+++ b/coding/python_style_guide.rst
@@ -235,8 +235,8 @@ Each Python file MUST contain the standard license preamble
 A copyright and license block using :ref:`the standard text <pkg-doc-code-preamble>` MUST be included at the top of each file.
 This can be written as a Python comment.
 
-.. literalinclude:: ../docs/snippets/license_preamble.txt
-   :language: text
+.. literalinclude:: ../docs/snippets/license_preamble.py
+   :language: python
 
 .. _style-guide-py-line-length:
 
--- a/docs/package_docs.rst
+++ b/docs/package_docs.rst
@@ -63,7 +63,7 @@ At the same time, READMEs aren't a primary source of Stack documentation.
 Thus we recommend you use your README to describe the purpose of the repository in slightly more detail than the GitHub summary line, and to point developers to the Stack documentation.
 At your discretion, READMEs can also be used for developer-centric notices that you feel shouldn't be included in the main documentation.
 
-To be consistent with the rest of the Stack's documentation, READMEs should be in reStructuredText format and named ``README.rst`` in the root of your repository.
+To be consistent with the rest of the Stack's documentation, READMEs should be in reStructuredText format and named :file:`README.rst` in the root of your repository.
 
 We suggest this template:
 
@@ -145,7 +145,7 @@ The institutions contributing to the Data Management software use the following 
 LICENSE
 -------
 
-A ``LICENSE`` file in the repository's root is the canonical description of LSST's code licensing.
+A :file:`LICENSE` file in the repository's root is the canonical description of LSST's code licensing.
 
 We use the GPLv3 license and a copy of it can be obtained from https://www.gnu.org/licenses/gpl.txt.
 
@@ -160,7 +160,7 @@ This should normally be included in a comment block at the top of the file and s
 .. literalinclude:: snippets/license_preamble.txt
    :language: text
 
-where the package name is stated explicitly at the top.
+where the package name is stated explicitly at the top to attach it to a particular :file:`COPYRIGHT` file.
 
 .. _pkg-doc-contributing:
 
@@ -168,7 +168,7 @@ CONTRIBUTING.rst
 ----------------
 
 `GitHub popularized the use of CONTRIBUTING files to help open source developers stay on the same page <https://github.com/blog/1184-contributing-guidelines>`_.
-Whenever a GitHub Issue or Pull Request is made, GitHub will display a link to the `CONTRIBUTING.rst` file.
+Whenever a GitHub Issue or Pull Request is made, GitHub will display a link to the :file:`CONTRIBUTING.rst` file.
 
 .. code-block:: rst
 
@@ -211,7 +211,7 @@ Whenever a GitHub Issue or Pull Request is made, GitHub will display a link to t
 
    .. _LSST DM Team Culture and Conduct Standards: https://confluence.lsstcorp.org/display/LDMDG/Team+Culture+and+Conduct+Standards
 
-.. note:: Some of the documentation URLs listed in this ``CONTRIBUTING.rst`` guide don't exist yet.
+.. note:: Some of the documentation URLs listed in this :file:`CONTRIBUTING.rst` guide don't exist yet.
 
 .. _pkg-doc-user-guide:
 
@@ -220,7 +220,7 @@ The Package's User Guide in docs/
 
 The heart of a Stack package's documentation are files in the ``docs/`` directory [#]_.
 This content is ingested by Sphinx, our documentation build tool, to publish user guides for each package.
-In the following section we describe how to write the main documentation file, ``docs/index.rst``.
+In the following section we describe how to write the main documentation file, :file:`docs/index.rst`.
 
 ..
    For complex packages, documentation can be split across many files in the docs/ directory.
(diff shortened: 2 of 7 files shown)


coding/python_style_guide.rst
```diff
@@ -227,6 +227,20 @@ Style changes must be encapsulated in a distinct commit (see :ref:`git-commit-or
...
+Each Python file MUST contain the standard license preamble
+-----------------------------------------------------------
+
+A copyright and license block using :ref:`the standard text <pkg-doc-code-preamble>` MUST be included at the top of each file.
+This can be written as a Python comment.
+
+   .. code-block:: python
+
+      #
+      # <template text>
```
parejkoj: Should something other than "<template text>" go in here? Is that supposed to be filled in automatically somehow?

docs/package_docs.rst
```diff
@@ -86,52 +86,101 @@ We suggest this template:
...
 ---------
 
-We assert copyright information in a centralized ``COPYRIGHT`` file, located in the repository's root.
+We assert copyright information in a centralized ``COPYRIGHT`` file, located in the repository's root rather than attempting to track copyright in each file separately within a package.
 Using a ``COPYRIGHT`` file allows us to maintain copyright information more effectively than in source code comments.
 (See `RFC-45 <https://jira.lsstcorp.org/browse/RFC-45>`_ for background on this policy).
 
-Default Copyright
-^^^^^^^^^^^^^^^^^
+Some code written before construction officially started will have an LSST Corporation copyright and this should not be changed unless it has erroneously been used for code after construction began in August 2015.
```
parejkoj: I would put this sentence about "old code" at the very end of this section. Otherwise, it's one of the first things a person sees.

coding/python_style_guide.rst
timj: I was trying not to duplicate the text from the other primary definition. I'm not sure if I can get sphinx to insert the text from a standalone text file and prefix each line with a comment character.

jonathansick: On the other hand, that boilerplate text shouldn't change without a big process (RFC, etc), so controlled duplication here is ok.

timj: It still seems fundamentally wrong to copy and paste it into three separate places.

jonathansick:
@timj would you mind moving all the unrelated changes to a different PR (even a user branch)? I think you can self-merge that stuff too. It would help the review.

coding/python_style_guide.rst
jonathansick: Okay, I'll look at how we can pull the content from here: https://github.com/lsst/templates/blob/master/file_templates/stack_license_py/example.py and https://github.com/lsst/templates/blob/master/file_templates/stack_license_cpp/example.cc

timj:
I wondered if you'd want me to self-merge all the cleanups first. I can do that.

docs/package_docs.rst
```diff
@@ -86,52 +86,101 @@ We suggest this template:
...
-
-The canonical source of this file's content is https://www.lsstcorp.org/LegalNotices/LsstLicenseStatement.txt, though ensure that it is named ``LICENSE`` in the repository.
-
 .. _pkg-doc-copyright:
 
 COPYRIGHT
 ---------
 
-We assert copyright information in a centralized ``COPYRIGHT`` file, located in the repository's root.
+We assert copyright information in a centralized ``COPYRIGHT`` file, located in the repository's root rather than attempting to track copyright in each file separately within a package.
```
jonathansick: Suggested edit:

```
We assert copyright information in a centralized ``COPYRIGHT`` file, located in the repository's root, instead of in individual source files.
```

The next sentence does the job of the "attempting to..." sentiment I think.

jonathansick: You could even slim down the next sentence to just

```
See `RFC-45 <https://jira.lsstcorp.org/browse/RFC-45>`_ for background on this policy.
```

docs/package_docs.rst
```diff
@@ -86,52 +86,101 @@ We suggest this template:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The institutions contributing to the Data Management software use the following copyright statements:
+
+* Association of Universities for Research in Astronomy
+* University of Washington
+* The Trustees of Princeton University
+* The Board of Trustees of the Leland Stanford Junior University, through SLAC National Accelerator Laboratory
+* University of Illinois Champaign-Urbana
+* California Institute of Technology
```
jonathansick: Could we add a sentence or two after this list to help contributors at other institutions? I'm thinking of contributors for places like the University of Pittsburgh or CC-IN2P3. Do they always get to add their institution to the copyright file? Or is it only institutions where the LSST Project has a formal intellectual property licensing contract?

docs/package_docs.rst
```diff
@@ -86,52 +86,101 @@ We suggest this template:
...
 
-   Copyright University of Washington and AURA/LSST (2012-2015)
+   Copyright 2012-2015 LSST Corporation
+   Copyright 2012-2015 University of Washington
+   Copyright 2015-2018 Association of Universities for Research in Astronomy
+
+These ``COPYRIGHT`` files should be updated during development and may be robotically refreshed.
+Automatically updating the files requires people committing to the repository use their :ref:`institutional email address <git-setup-institutional-email>`.
+
+The downside of using a single ``COPYRIGHT`` file is that moving a file between packages requires that both the source and destination ``COPYRIGHT`` files may need to be amended, since the copyright information is not included in the file being moved.
```
jonathansick: Instead of starting the sentence with "The downside [..]", which makes it sound like an editorial comment, let's try to start with a task prompt. For example:

```
If you move source files between packages, remember to audit the ``COPYRIGHT`` file and move copyright assignments between packages as necessary.
```

docs/package_docs.rst
timj: @ktlim has previously expressed a desire not to use RFCs as part of the documentation if we can avoid it. I tend to agree that RFC-45 is not a coherent view of policy. What it shows is a lot of angst.

docs/package_docs.rst
timj: They should add their own institution if the contribution is significant. What I have learned from reading the GPL help guide is that adding or editing a few lines does not count as being enough to confer copyright.

docs/package_docs.rst
jonathansick: Sounds good. I'd just avoid the word "attempting" since it sets the wrong tone.

timj:
I think I've dealt with the comments. The remaining issue is how to generate the preamble text from the source of truth in the templates repo.

timj:
@jonathansick I think for the next steps, once you are okay with what I've written, I'd like to get this formally reviewed by Wil and John. At that point we go live with the developer guide and update the templates package to match (and do auto updating and the like).

docs/snippets/license_preamble.txt
```diff
@@ -0,0 +1,20 @@
...
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
timj: I've removed the LSST link from here because it is confusingly different to an actual GPL license and includes every license that LSST has found in code it uses. An option is to add the link to the root of the LSST licenses directory to the top of the license text associated with the lsst.org link.

docs/snippets/license_preamble.txt
```diff
@@ -0,0 +1,20 @@
+This file is part of %package_name%.
+
+Developed for the LSST Data Management System.
+This product includes software developed by the LSST Project
+(http://www.lsst.org).
+See the COPYRIGHT file at the top-level directory of this distribution
+for details of code ownership.
+
+This program is free software: you can redistribute it and/or modify
```
timj: GNU recommends we replace "This program" with the actual package name in the preamble. I'm happy to do that, especially if the templates builder can fill it in.

docs/snippets/license_preamble.txt
```diff
@@ -0,0 +1,20 @@
+This file is part of %package_name%.
+
+Developed for the LSST Data Management System.
+This product includes software developed by the LSST Project
+(http://www.lsst.org).
+See the COPYRIGHT file at the top-level directory of this distribution
```
timj: There does seem to be an option to add a line here along the lines of:
```
Copyright LSST Data Management Contributing Institutions.
See the COPYRIGHT...
```
It's not a legal statement but it does give the reader a hint of what will be found in the COPYRIGHT file.

docs/snippets/license_preamble.txt
jonathansick: This is a pattern I see in many open source projects: https://github.com/astropy/astropy/tree/master/licenses But I can't personally comment on what we should actually do here. Having the real GPLv3 license as our own license text seems like a good idea, though.

jonathansick:
@timj Feel free to elevate the review. Text looks good, as far as I can review. I'm not entirely clear on the automation strategy given https://github.com/templates but I'm not personally worried about that yet.

docs/snippets/license_preamble.txt
timj: That pattern seems fine  of putting all the myriad sublicenses in a subdirectory but having the proper license at the top level. The LSST "license" combined them all into a single confusing file which Github then has no idea what to do with. My other comment from that link is that I had no idea we have an AURA license.

timj:
@womullan according to the link to the lawyer I included and GPL you need to include the publication year. 2015 is not sufficient if you have published additional content in 2017. I am assuming that for us we publish the code weekly so we don't really have to think about it and just add the current year when we make a substantive change.

womullan:
@timj According to other sources including wikipeda the copyright notice is  informational so its form is not so important. Techhically I suppose you are correct it should carry the year of first publication which would exted to 2032 (?)  for us.  I do not mind really.

Wikipeda says" US law no longer requires the use of a copyright notice, although placing it on a work does confer certain benefits to the copyright holder. Prior law did, however, require a notice, and the use of a notice is still relevant to the copyright status of older works.

For works first published on or after March 1, 1989, use of the copyright notice is optional. "

timj:
Yes, @frossie has been noting for a while that an explicit copyright notice is not required, given we have source code history in git. It's the polite thing to do though since it makes it easier for the casual user of the package (although separating copyright information from the source files makes it harder to keep track of what's going on if a third party integrates some of our code into their package).

@womullan did you have thoughts on whether "This Program" in the license preamble should be replaced with the package name as recommended by GNU?

@jonathansick what do you want to do about the first note in the file which tells people these instructions should be ignored?

womullan:
@timj  thought you were already going to put "package name" - it is better.

timj:
Yes, there has to be some statement of package name in the preamble and we do have that. I was wondering about the second two "This program" lines. I'm okay with leaving that as is.

timj:
I've spoken to @jonathansick and he will sort out the note at the top of the page in time for AHM in March.

PR #209, opened 2018-03-16
DM-13656: Major re-organization

jonathansick:
Draft: https://developer.lsst.io/v/DM-13656

This PR implements the reorganization proposed in [RFC-453](https://jira.lsstcorp.org/browse/RFC-453).

## About the commit structure

The commits are generally structured in three parts:

1. Move/rename pages without changing the content
2. Update cross-links
3. Add redirect pages at old URLs

Doing this provides a more usable Git history where we can see how the page was moved. We can also see where redirect pages were added so that they can be replaced by an automated process later on.

## About the table of contents

The table of contents management is more complicated than it used to be, for the sake of usability. I think that it can be simplified in the long run.

- The `toctrees` with labels are needed by this particular theme to populate the side bar. Most themes don't use toctree labels to manage headers in the sidebar. So once we switch to a new, custom, LSST theme we can move away from this.
- The lists of `doc` roles in `index.rst` provide the flexibility to group topics and include sub-headers. Again, once we switch to a new theme, this can be consolidated with regular `toctrees`
- The `index.rst` files in individual subdirectories mirror the consolidated contents list in the root `index.rst`. This is needed so that users who manually edit URLs down to the directory path don't get a 404. At the same time, this isn't the primary navigation because I want most pages to be accessible from the guide's homepage. Eventually, I think I can make some tooling to automatically generate these directory listings. This is a pattern that might come up with our other user guides too.

## Next steps

This PR puts the Developer Guide in a more usable state. The next steps, in subsequent PRs, is to go into the content and re-write/edit it for more usability. For example:

- Break up the `/work/flow.html` page so that it serves as a compact overview. Additional details and policies would go into `/git/` and `/github/` sections.
- Add a `/github/` section that describes how we use GitHub (PRs, orgs, teams, etc.)
- Add an overview section at the top of the "Services" part, which would include an overview of how ot get help with different types of DM services, and include a page or two on the ITSC.

****

screenshot-2018-3-16 lsst dm developer guide lsst dm developer guide dm-13656 documentation


kfindeisen:
I think I can find everything I normally want to look up. :+1:

communications/rfc.rst
```diff
@@ -0,0 +1,115 @@
...
+
+.. _rfc-cases:
+
+Cases that require an RFC
+=========================
+
+Some cases **always require** an RFC:
+
+- :doc:`/build-ci/third_party`.
+- :ref:`Releasing a change-controlled document <ccd-release>`.
```
ktlim: Also `/stack/adding-a-new-package`.

communications/rfc.rst
```diff
@@ -0,0 +1,115 @@
...
+Those stories must have an **Is triggered by** relationship with the parent RFC issue.
+
+Next, see :ref:`rfc-implementing`.
+
+.. _rfc-escalating:
+
+Escalating an RFC
+-----------------
+
+If the discussion is not converging, but you still feel that a decision must be taken, you (or the DMCCB) may escalate the RFC to be considered by the DM Configuration Control Board (DMCCB).
```
ktlim: You don't say how.  "Click the **Escalate RFC** button on the RFC's JIRA issue page."

communications/rfc.rst
```diff
@@ -0,0 +1,115 @@
...
+Some types of RFCs are immediately escalated as a matter of routine, see :ldm:`294` C.3.
+
+.. _rfc-widthraw:
```
ktlim: "withdrawing" (for consistency)

communications/rfd.rst
```diff
@@ -0,0 +1,63 @@
...
+   This page is **descriptive**.
+   See :ldm:`294` C for normative documentation of DM's discussion and decision making process.
+
+.. See RFC-53: https://jira.lsstcorp.org/browse/RFC-53
+
+############################
+Request for discussion (RFD)
+############################
+
+**Request for discussion (RFD)** is a process to facilitate in-depth technical discussions across DM.
```
ktlim: An RFD is a thing, not a process.  The RFD Process is a process.

communications/rfd.rst
```diff
@@ -0,0 +1,63 @@
...
+These might be:
+
+- Detailed design discussions for a component of the system or its interfaces.
+- Design reviews for new code or refactorings of old code.
+- Brainstorming methods to solve difficult problems.
+- "Brain dump" explanations of a design to share knowledge across DM.
+
+.. seealso::
+
+   :doc:`rfc` is an additional process you can use to seek adoption of a specific proposal.
```
ktlim: "can" -> "should"

communications/rfd.rst
```diff
@@ -0,0 +1,63 @@
...
+Creating an RFD
+===============
+
+Create an RFD by `making a post`_ on Community in the  `DM RFD category`_.
+A `template is provided in the post <https://community.lsst.org/new-topic?category=data%20management/dm%20rfd>`_ to allow you to suggest the date and time of the discussion (:ref:`see below for time slot <rfd-time>`), the medium for the discussion, and to specify the suggested audience.
+In the RFD's description:
+
+- Summarize the issue, and indicate a desired outcome from the discussions.
+- Include background material (using attachments, if necessary).
+- Provide a link to the BlueJeans or Google Hangouts room.
```
ktlim: I think "or Google Hangouts" can probably be deleted now.

communications/rfc.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Cases that require an RFC
+=========================
+
+Some cases **always require** an RFC:
+
+- :doc:`/build-ci/third_party`.
+- :ref:`Releasing a change-controlled document <ccd-release>`.
+
+.. _`#dm`: https://lsstc.slack.com/messages/C2JPL2DGD
```
ktlim: How come this URL is different (and less readable) than the one in `rfd.rst`?

communications/rfd.rst
```diff
@@ -0,0 +1,63 @@
...
+.. _`DM RFD category`: https://community.lsst.org/c/dm/dm-rfd
+.. _`dm-devel mailing list`: https://lists.lsst.org/mailman/listinfo/dm-devel
+.. _`'#dm' Slack room`: https://lsstc.slack.com/messages/dm/
+
+.. _rfd-time:
+
+The RFD time slot
+=================
+
+RFDs can scheduled for any convenient time, but we do have a weekly reserved time slot on Tuesdays from 12:30 to 2 PM Pacific.
```
ktlim: Needs a "be" after "can".

communications/rfd.rst
```diff
@@ -0,0 +1,63 @@
...
+
+If there are no requests 24 hours before a given time slot, the meeting will be canceled and the time freed up for other activities.
+
+If there are conflicting claims to the RFD time slot, the System Architect will arbitrate.
```
ktlim: Not sure who wants this responsibility now that that title has been retired, although we haven't had any RFDs lately.

LICENSE
```diff
@@ -1,3 +1,395 @@
...
+     rights in the material. A licensor may make special requests,
+     such as asking that all changes be marked or described.
+     Although not required by our licenses, you are encouraged to
+     respect those requests where reasonable. More_considerations
+     for the public:
+	wiki.creativecommons.org/Considerations_for_licensees
+
+=======================================================================
+
+Creative Commons Attribution 4.0 International Public License
```
ktlim: I think the actual license starts here; the rest of the prefatory stuff needs to be removed.

LICENSE
```diff
@@ -1,3 +1,395 @@
...
+     failure to comply consented to unless expressly agreed to by the
+     Licensor.
+
+  d. Nothing in this Public License constitutes or may be interpreted
+     as a limitation upon, or waiver of, any privileges and immunities
+     that apply to the Licensor or You, including from the legal
+     processes of any jurisdiction or authority.
+
+
+=======================================================================
```
ktlim: And delete from here down.

python/profiling.rst
```diff
@@ -0,0 +1,118 @@
...
+If you do:
+
+.. code-block:: python
+
+   p.print_callees("readCatalog")
+   p.print_callees("getAllForCcd")
+
+you can see that the **cumtime** column doesn't add up to the **cumtime** values in the above, so the remaining time is time spent within those functions doing work.
+
+For more details on pstats and python profiling in general see http://docs.python.org/2/library/profile.html.
```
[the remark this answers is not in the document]
timj: You can delete the `/2/` and it will do the right thing and go to py3.

LICENSE
jonathansick: I still think this is correct. Licenses are detected by [licensee](https://github.com/benbalter/licensee):

1. [licensee docs say it uses choosealicense.com as its source](https://github.com/benbalter/licensee/blob/master/docs/what-we-look-at.md#known-licenses).
2. choosealicense's [text for the CC-BY 4.0 license](https://choosealicense.com/licenses/cc-by-4.0/) is exactly what I've included in LICENSE, according to their instructions.

I'd like to stick with this as-is.

ktlim: But the text itself says "The following considerations are for informational purposes only, are not exhaustive, and do not form part of our licenses."  How can we include text that is not part of the license as part of the license?

ktlim: Maybe this is a case where we have to fix upstream...

jonathansick:
Thanks @ktlim, I think I've addressed all the feedback. Additional comments:

- I'd like to leave the LICENSE file as-is and see if it works properly on GitHub
- I'd like to hold off on putting an example notation right at the top of the Developer Guide homepage because I think this will cause some design issues. I can create a new derivative ticket to standardize our annotations of normative pages, and add a help page related to this.
- I've added the full text from LDM-294 Appendix C back into the RFC page.

jonathansick:
Moving the [line conversation about LICENSE](https://github.com/lsst-dm/dm_dev_guide/pull/209#discussion_r175240007) back to the top-level since it seems to be the overall blocker right now.

- I've confirmed by running [licensee](https://github.com/benbalter/licensee) that licensee will not detect a CC-BY 4.0 license text that is missing the non-legal code parts of the CC-BY 4.0 license.
- I've found a [page that describes the CC licenses](https://wiki.creativecommons.org/wiki/Legal_code).
  > Every Creative Commons license has three layers: the lawyer-readable legal code, the human-readable deed, and the machine-readable code. As the only legally-operative layer, the legal code is the primary layer of the CC licenses. It consists of the text of the licenses, as well as any meaningful formatting and other inseparable elements listed below.

  I think what this is saying is that the CC-BY *license* contains non-legally binding elements, but those elements are still part of the license document.
- I've found this [GitHub conversation by someone who purports to represent CC](https://github.com/github/choosealicense.com/issues/33#issuecomment-33057403):
  > To clarify, the preamble and disclaimer are not part of the license--they are included in the license files CC publishes, but while CC has committed not to change the legal code itself after publication, this material may change. (It is also permissible to omit them, but we recommend against it.) What is and isn't part of the license is noted at http://wiki.creativecommons.org/Legal_code for reference.

  The operative aspect here is that CC **recommends against** removing the preamble and disclaimer from the license document.

Thus I stand firm that the treatment of the LICENSE file in this PR is proper and following best practice.

ktlim:
Thanks for finding those references and testing things; serves me right for trying to apply logic to the law.  I still see two minor issues:

- Under `communications/rfc.html#extended-discussion-of-the-rfc-process`, "If the three criteria set in Empowerment of DM team members" is missing "are not met".
- `build-ci/new_package.html` says "Before adding a new dependency to any of these products, it must be approved through the RFC process," so it should be added to the list in `communications/rfc.html#standard-procedures-that-require-an-rfc`

jonathansick:
Good catches! Thanks @ktlim

PR #210, opened 2018-03-20
DM-12571 Replace super() section with new text.

parejkoj:


coding/python_style_guide.rst
```diff
@@ -619,21 +619,23 @@ Within a module, follow the order:
...
+As a result, there are a few argument-related caveats about the use of ``super()`` in multiple inheritance hierarchies:
 
-- Never call :py:func:`super` with anything but the exact arguments you received, unless you really know what you're doing.
-- When you use it on methods whose acceptable arguments can be altered on a subclass via addition of more optional arguments, always accept ``*args, **kw``, and call :py:func:`super` like ``super(MyClass, self).currentmethod(alltheargsideclared, *args, **kwargs)``.
-  If you don't do this, forbid addition of optional arguments in subclasses.
-- Never use positional arguments in ``__init__`` or ``__new__``.
-  Always use keyword args, and always call them as keywords, and always pass all keywords on to :py:func:`super`.
+* Only pass ``super()`` the exact arguments you received.
+* When you use it on methods whose acceptable arguments can be altered on a subclass via addition of more optional arguments, always accept ``*args``, ``**kw``, and call ``super()`` like ``super().currentmethod(alltheargsideclared, *args, **kwargs)``.If you don’t do this, forbid addition of optional arguments in subclasses.
```
ktlim: Need a space after the period.

coding/python_style_guide.rst
```diff
@@ -619,21 +619,23 @@ Within a module, follow the order:
...
+* Never use positional arguments in ``__init__`` or ``__new__``.  Always use keyword args, and always call them as keywords, and always pass all keywords on, e.g. ``super().__init__(**kwargs)``.
```
ktlim: Maybe "call them using keywords"?

coding/python_style_guide.rst
```diff
@@ -619,21 +619,23 @@ Within a module, follow the order:
 
 .. _style-guide-py-super:
 
-``super`` SHOULD NOT be used unless the author really understands the implications (e.g. in a well-understood multiple inheritance hierarchy).
-----------------------------------------------------------------------------------------------------------------------------------------------
+``super`` MAY be used with care to call inherited methods.
+----------------------------------------------------------
 
-Python provides :py:func:`super` so that each parent class' method is only called once.
+Python provides ``super()`` so that each base class’s method is only called once.
```
ktlim: Does ``:py:func:`super` `` produce better output, semantically even if not visually?

parejkoj: Nice! That actually links it. Thanks for that one.

coding/python_style_guide.rst
```diff
@@ -2,7 +2,7 @@
 DM Python Style Guide
 #####################
 
-This is the version 6.0 of the DM Python Coding Standard.
+` is the version 6.0 of the DM Python Coding Standard.
```
ktlim: What happened here?

parejkoj: Oops. Good catch.

timj: Are we updating that number ever? I don't remember updating it when I've made previous changes.

jonathansick: It's a hold-over from the Confluence page. Maybe A subject for another PR, but I'd say we should either drop it (the only standard that matters is always the current standard), or get in the matter of maintaining a change log if we actually do care about versioning the style guide.

parejkoj: Good point. I'm not going to touch it now (the above was a mistake on my part), but we should probably remove it.

parejkoj:
I've pushed an update based off of a discussion with Colin, who wanted a concrete example of what to do. Comments welcome.

PR #212, opened 2018-04-05
DM-14014: Remove "Keyword assignment operators SHOULD be surrounded by a space..." from the Python style guide

r-owen:


timj:
I think E251 is also in tools/sublime.rst

PR #213, opened 2018-04-06
add travis shellcheck configuration examples

jhoblitt:
Due to that fact that many EUPS products for 3rd party code include at least a `ups/eupspkg.cfg.sh` shell script, it is reasonable to enable `shellcheck` linting by default.

PR #215, opened 2018-04-12
disable travis email + restrict branch builds

jhoblitt:


jhoblitt:
Wait, something funky happened with the rebase.

jhoblitt:
fixed.

jonathansick:
If someone pushes a branch to GitHub, but does not create a PR, will no tests be run until either

- the PR is created
- the branch is merged to `master`

Is that right?

jhoblitt:
That's correct. AFAIK there is no way to automatically cancel a branch build when an associated PR is opened.

jhoblitt:
(or the branch matches the regex, eg, `15.0`)

jonathansick:
I'm not comfortable with that loophole. If we really want to eliminate the double build, I think we should be eliminating the PR test, and focus on the push-based testing instead. Disabling the ["Build pushed pull requests" button](https://docs.travis-ci.com/user/web-ui/#Build-pushed-pull-requests) would do that.

jhoblitt:
How is it a loophole? The PR is CI'd before merging -- just as it would be if the PR originated from a branch in a fork.

Further, branch and PR builds are not equivalent.  A branch is built unmodified while a PR is merged into the target branch before being built (this is why the sha1s on PR builds don't match the tip of the PR).

In any event, not a big deal but it building twice contributes to scheduling latency for the entire org.

jonathansick:
The loophole I'm referring to is that a branch can be merged untested without a PR according to my original scenario. Yes, its our standard, but sometimes "accidents" happen, and the point of CI is to protect `master` from such accidents.

On the other hand, with GitHub branch protections, do we know if "Require status checks to pass before merging" will work to close the loophole? I.e., will it require an affirmative Travis check and that can only happen if the branch is subject to a PR?

jhoblitt:
Branch protection would be pretty useless if that wasn't the case. 

> 
> Require status checks to pass before merging
> 
> Choose which status checks must pass before branches can be merged into master. When enabled, commits must first be pushed to another branch, then merged or pushed directly to master after status checks have passed.

stack/examples/flake8-travis.yml
```diff
@@ -6,3 +6,9 @@ matrix:
       install:
       - pip install flake8
       script: flake8
+branches:
+  only:
```
jonathansick: Please add a comment here explaining the rationale for this. E.g.

```
# Only build on pushes against these branches (avoids double-testing PRs)
```

ktlim:
After a little further consideration of #213 and this PR, I'm reminded of @jdswinbank 's request that developers not have to think about this.  I think that means that these examples should be removed from the dev guide, which has the benefit that they don't have to be maintained (and I don't have to review their updates!).  What do you folks think?

timj:
I thought from the Slack discussion that John was happy with people copying in the travis file to the repo so long as the developers didn't have to fiddle with the GitHub settings. If that is so then we need the examples to be somewhere (templates instead of dev guide?)

jhoblitt:
Per discussion, this PR is being retracted.

PR #220, opened 2018-04-24
Update data_protection.rst

mfisherlevine:
Add some instructions on how to do self-service data retrieval.

### DIFF
--- a/services/data_protection.rst
+++ b/services/data_protection.rst
@@ -65,6 +65,10 @@ Protection Capabilities
     Data locations supporting this tier of protection allow for users to retrieve,
     at their discretion and leisure, a previous data copy limited to a maximum age
     of 30 days. Data set versions are captured at a maximum of one per day.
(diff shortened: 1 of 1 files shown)


services/data_protection.rst
```diff
@@ -65,6 +65,10 @@ Protection Capabilities
     Data locations supporting this tier of protection allow for users to retrieve,
     at their discretion and leisure, a previous data copy limited to a maximum age
     of 30 days. Data set versions are captured at a maximum of one per day.
+    
+    Those wanting to use this self-serve data retrieval service can go to:
+    ``home/.snapshots/home_YYYYMMDD_000N/<username>/``
```
ktlim: Put a "/" before the first "home". I think N will always be 1.  And you should indicate that this also works for the other filesystems with the exception of "/scratch".  Otherwise, you're OK to merge.

PR #225, opened 2018-05-02
Add a paragraph on Python exceptions.

ktlim:


parejkoj:
I'd move the sentence about using the appropriate standard Python exception to the top: that's the thing we want to emphasize. Most of the time you don't need to define your own exception.

The final sentence should say: ``` ``pex_exceptions`` should not be caught unless there is no equivalent python binding (e.g. lsst.pex.exceptions.LengthError).```

parejkoj:
Oh, and please link the statement about defining a module-specific exception to the python docs. This might be a good link: https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions

ktlim:
I want people to at least consider a module-specific exception each and every time.  I think they're underused (perhaps not massively, but at least somewhat).

Saying "should not be caught" has the wrong implication to me: that they should be allowed to pass through to higher-level code for some reason.  But the point about stating this in terms of C++ exceptions that have no Python equivalent is a good one.

I'll edit later tonight.

ktlim:
Is that a bit better, @parejkoj ?

ktlim:
Self-approved and merged.

PR #226, opened 2018-05-03
DM-14299: Remove Python 2 from developer guide

timj:


python/style.rst
```diff
@@ -24,30 +24,9 @@ All DM Python code MUST work with Python 3
 ------------------------------------------
 
 All the Python code written by LSST Data Management must be runnable using Python 3.
-Python 2 will cease to be supported before LSST is operational (:pep:`373`).
+Python 2 is not supported.
```
ktlim: Maybe "is not required to be" instead of "is not"?

python/style.rst
```diff
@@ -562,13 +541,12 @@ Within a module, follow the order:
...
 2. Module-level comments (such as the `license statement <https://github.com/lsst/templates/tree/master/file_templates/stack_license_py>`__)
 3. Module-level docstring
-4. ``from __future__ import absolute_import, division, print_function``
-5. ``__all__ = [...]`` statement, if present
-6. Imports (except ``from __future__ import...``)
-7. Private module variables (names start with underscore)
-8. Private module functions and classes (names start with underscore)
-9. Public module variables
-10. Public functions and classes
+4. ``__all__ = [...]`` statement, if present
```
ktlim: Not that it's a big deal, but is there an ReST way to do this so we don't have to manually renumber?

python/style.rst
```diff
@@ -677,46 +655,9 @@ Writing Python with accepted patterns makes your code easier for others to under
 
 `Fluent Python <http://shop.oreilly.com/product/0636920032519.do>`_ by Luciano Ramalho is an excellent guide to writing idiomatic Python.
 
-Idiomatic Python also reduces technical debt, particularly by easing the migration from Python 2.7 to Python 3.
-Codes should be written in a way that helps the futurize_ code converter produce more efficient code.
+Idiomatic Python also reduces technical debt.
 For more information see the online book `Supporting Python 3 <http://python3porting.com/toc.html>`_ by Lennart Regebro.
 
-.. _futurize: http://python-future.org/futurize.html
```
ktlim: Removing this is OK, although I wonder if it should be left somewhere as "if you're going to go to the trouble of supporting Py2, here's how to do it".

PR #230, opened 2018-05-30
Update SublimeLinter configuration for SL4

parejkoj:
SublimeLinter4 changed its configuration, and I've finally sorted out how it works. I've opened an upstream PR with SL-flake8 to update their readme:

https://github.com/SublimeLinter/SublimeLinter-flake8/issues/96

Fortunately the new SL4 settings don't require having every setting repeated in them, so the settings file is much shorter.


editors/sublime.rst
```diff
@@ -142,70 +142,23 @@ You can get SublimeText to check your python code inline and mark lines that do
 Install ``SublimeLinter`` and ``SublimeLinter-flake8`` via Package Control.
 Use the following configuration to conform to LSST's python style, to mark failing lines, and to provide a summary of failures on save that will let you go directly to those lines.
 
-
-You may have to `configure your PATH <http://www.sublimelinter.com/en/latest/usage.html#how-linter-executables-are-located>`_ to allow SublimeText to find the ``flake8`` executable.
-
+Try starting SublimeText with the `subl command <https://www.sublimetext.com/docs/3/osx_command_line.html>`_ after setting up e.g. `lsst_distrib`, to get sublime to use the stack-installed version of `flake8`. If it does not work that way, you may have to `configure your PATH <http://www.sublimelinter.com/en/latest/usage.html#how-linter-executables-are-located>`_ to allow SublimeText to find the ``flake8`` executable.
```
r-owen: I'm not happy with this wording. What is important here is that if a user starts ST by clicking on the icon, or does not setup the LSST software before running it from the command line, then ST will not be able to find flake8. You really should emphasize that. Personally I think it safer and more robust to install my own flake8 in my default Python. Then ST always finds it.

I also suggest that you have users setup a default flake8 configuration in the standard location (~/.config/flake8) as this will be used by ST, flake8 on the command line and probably basically any other flake8 linter (e.g. in Atom).

Here is possible wording for both of these:

Save the following as `~/.config/flake8` to provide a default configuration for `flake8`:

    [flake8]
    # from https://developer.lsst.io/python/testing.html#enabling-additional-pytest-options-flake8
    ignore = E133, E226, E228, N802, N803, N806
    max-line-length = 110

There are several ways to make Sublime Text find flake8:
- Install flake8 in your default Python. This is simple and reliable, but requires maintaining your own copy of flake8 in addition to the version provided by the LSST DM stack.
- To use the standard LSST flake8 you can specify the path to that flake8 in Sublime Linter settings; however, that path may change if if the LSST version changes. Sublime Text can find the LSST version of flake8 automatically provided you always start Sublime Text from the command line and that you setup your LSST software before starting Sublime Text.

parejkoj: If you want to recommend that people install their own flake8, we should get some input from @timj , since pip installing packages that exist in eups can cause trouble.

timj: For Atom, I explicitly point it at my `/usr/local` python because on a Mac when an App starts it doesn't necessarily pick up your shell environment and definitely won't be guaranteed to pick up an eups environment (so the lsstsw flake8 eups package won't ever be seen). You need to make sure it picks up a python3 flake8 so that it doesn't get confused by `f`-strings and the like. I also want Atom to still work with flake8 even if I am reinstalling my lsstsw environment.

parejkoj: Could someone please provide the exact text they'd prefer for that configuration (and maybe have a person go and try said instructions)? It sounds like your suggestion @timj involves people installing a `brew` python and using its `pip` to install `flake8`.

I'm also not sure about recommending that people set a global flake8 configuration. If so, we should probably write those instructions somewhere else and link them from here (for those who don't want to configure just SublimeText's flake8 linter).

timj: I'm not sure I can insist that people use a brew python flake8 or any other. It's probably a bit easier to use the conda python now that we don't have the risk of python2 flake8 and python3 flake8 giving different answers. I do think that people do need to install their own flake8 from, at minimum their lsst anaconda.

morriscb:
@parejkoj asked me to review this as well. From a readability stand point it's not clear where to put the text from SublimeLinter code block (I initially put it in the generic settings). Some text describing what drop down menu to edit it in would be great. Also he told me when I pointed it out, the second block of code in the Linter section doesn't make sense any more and he has removed it. Finally there was a typo in the SublimeText settings where a comma was missing.

parejkoj:
I've updated it and @morriscb tried it out successfully. @r-owen : can you please sign off, if you're happy with this version?

r-owen:
This looks a lot better. I'm puzzled why you find the flake args necessary. I don't use them and have no problems. It is either picking up my `~/.config/flake8` file or the settings.cfg -- I'm not sure which. But either is a lot better than hard coding args in my opinion since they are the settings that flake8 will use on the command line and in other editors. Note that we now disable a additional warning if we are using new enough flake8 (I have not had to do this yet).

PR #232, opened 2018-06-14
DM-9790: Items missing from the pybind11 wrapping guide

r-owen:


pybind11/how-to.rst
```diff
@@ -461,6 +467,39 @@ The end result of all the steps above looks like this:
...
+
+The Python name for your wrapper module is `exampleOne`.
+If the wrapped classes can be returned by a function or unpickled then it is crucial that your module is imported when the package is imported.
+If the symbols are part of the public API then this is typically done by adding the following to your package's main ``__init__.py`` file:
+
+.. code-block:: python
+
+    from exampleOne import *
+
+If you don't want your wrapper's symbols in your package's top-level namespace then you can use ``from . import exampleOne`` or import the module in some other way.
```
parejkoj: "or import the module in some other way" - I have no idea what the other options would be.

pybind11/how-to.rst
```diff
@@ -220,6 +220,12 @@ We wrap the class using the ``py::class_<T>`` template:
 
     As in the example, classes should almost always have a :ref:`shared_ptr holder type <style-guide-pybind11-holder-type>`.
 
+.. note::
+
+    It is far more common to name the pybind11 class ``cls`` and wrap one class per module or per function within a module.
+    However, the examples in this document are quite distributed, so to make them easier to follow we use ``clsExampleOne``.
+    See also :ref:`class prefix <style-guide-pybind11-class-prefix>`.
```
parejkoj: Hmm... I'm not sure that this clarifies things. The style guide doesn't say anything about "one class per function": it seems to imply that wrapping more than one class per file means you *have* to use `clsExample`.

r-owen: I looked at the style guide entry and I think `wrapper` in this case can refer to either a function or the full wrapper, as appropriate. The text should be cleaned up to clarify that. I suppose I could do that on this ticket, though I wonder if that needs an RFC.

parejkoj: I also do one class per function, as it lets you group similar classes in one file, but keep them distinct.

Probably doesn't need an RFC and do it on this ticket. Give it to Pim or Jim or Krzysztof to review, since they're the ones who would care the most. A brief review of our existing code suggests its what we're doing already in many places.

pybind11/how-to.rst
parejkoj: Following "explicit is better than implicit", I'd suggest stopping the sentence at `exampleOne`. I think we're much better off having two explicit options (`from X import *` and `from . import X`), and not implying the others, which are too implicit and "magic" to me.

parejkoj:
Thanks for updating that. I guess @ktlim should probably sign off on the style guide changes, to make it official. But I'm happy with them.

PR #243, opened 2018-07-20
Add copyright generation script.

ktlim:
Also fix some incorrect text in the copyright overview.


timj:
See also #242 which is also attempting to fix the construction start date.

ktlim:
Rebased on DM-15172.

PR #260, opened 2018-09-12
Smaller example repo

frossie:
So testdata_decam is 1.6 GB so takes a while. Testdata_subaru is 190 MB and that is also a git-lfs repo so maybe better for a test


frossie:
Feel free to reject, just a suggestion.

PR #262, opened 2018-09-21
DM-10384: rewrite pybind11 how-to and style guide

TallJimbo:


pybind11/how-to.rst
```diff
@@ -166,82 +166,72 @@ Adding dependencies
...
-    }}}  // lsst::tmpl
+    PYBIND11_MODULE(_tmpl, mod) {
+        utils::python::WrapperCollection wrappers(mod, "lsst.tmpl");
+        // wrapper code goes here
+        wrappers.finish();
+    }
+
+    }}  // lsst::tmpl
+
+``WrapperCollection`` is a helper class provided by the LSST ``utils`` package that should be used in most LSST pybind11 wrappers.
```
ktlim: Is there ever a case when it shouldn't be used?

pybind11/how-to.rst
```diff
@@ -166,82 +166,72 @@ Adding dependencies
...
-.. note::
-
-    Add ``.export_values()`` if (and only if) you need to export the values into the
-    class scope (so they can be reached as ``ExampleOne.RED``, in addition to ``ExampleOne.State.Red``).
+``WrapperCollection`` will automatically call the callback given as the second argument with the module and the ``pybind11::class_`` object passed as the first argument, but not until all other classes defined in the module have been declared.
+That ensures all types are known to pybind11 before any signatures are declared, which in turn ensures pybind11 can generate the right type conversions when wrapping those signatures.
 
-    Never do this for new style scoped ``enum class`` types, since that will give
-    them different symantics in C++ and Python.
+Usually the wrappers for a class can be defined entirely within the callback function, but ``wrapType`` also returns the ``class_`` object in case its needed elsewhere.
```
ktlim: "its" -> "it's"

pybind11/how-to.rst
```diff
@@ -166,82 +166,72 @@ Adding dependencies
...
+where ``{{pkg}}`` is the name of the package, you will need to add the dependency ``setupRequired(pybind11)``.
 You also need to modify the ``dependencies`` in ``{{pkg}}/ups/{{pkg}}.cfg``, by adding ``"pybind11"`` to ``"buildRequired"``.
 
 .. _creating-module-file:
 
 Creating a module file
 ^^^^^^^^^^^^^^^^^^^^^^
 
-Following :ref:`our rules on file naming <style-guide-pybind11-module-naming>`, we start by creating a minimal module file ``python/lsst/TMPL/exampleOne.cc`` with the following content:
+Following :ref:`our rules on file naming <style-guide-pybind11-module-naming>`, we start by creating a minimal module file ``python/lsst/TMPL/_tmpl.cc`` with the following content:
```
ktlim: Wouldn't it be better to be consistent about one-cc-file-per-include, instead _adding_ a `_tmpl.cc` that calls the `wrapExampleOne` function?

pybind11/how-to.rst
```diff
@@ -292,32 +282,37 @@ Wrapping a member function is easy:
...
+Following the :ref:`rule on STL containers <style-guide-pybind11-stl-containers>` we simply include the ``pybind11/stl.h`` header (to enable automatic conversion to and from Python containers) and wrap the function as normal:
 
 .. code-block:: cpp
 
-    clsExampleOne.def("computeSomeVector", &ExampleOne::computeSomeVector);
+    cls.def("computeSomeVector", &ExampleOne::computeSomeVector);
+
+.. note::
+
+    The converters defined in ``pybind11/stl.h`` do a conmplete conversion
```
ktlim: "conmplete" -> "complete"

pybind11/how-to.rst
```diff
@@ -353,28 +348,47 @@ Wrapping *static* member functions is trivial:
...
 .. note::
 
     * We use ``py::is_operator()`` to return ``NotImplemented`` on failure.
     * We don't use ``py::is_operator()`` for in-place operators as this can lead to confusing behavior.
     * We name the lambda arguments :ref:`self <style-guide-pybind11-lambda-self-argument>` and :ref:`other <style-guide-pybind11-lambda-other-argument>`.
 
+Module-Level Declaration
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Module-level free functions and variables can be declared inside a ``wrapType`` callback, and should when they're closely related to the class it defines.
```
ktlim: ", and you should do so when they're closely related to the class they define"

pybind11/how-to.rst
```diff
@@ -353,28 +348,47 @@ Wrapping *static* member functions is trivial:
...
+    wrappers.wrapFunctions(
+        [](auto & mod) {
+            // any number of module-level wrappers go here
+        }
+    );
+
+.. note::
+
+    We do not attempt to update the ``__module__`` attribute of free functions, as these are rarely used.
+    Free functions for which ``__module__`` is important (e.g. functions passed to ``__reduce__`` for pickling) should have their ``__module__``modified manually.
```
ktlim: Space before "modified".  I'm not sure an "e.g." is sufficient for someone to know when this is important.

pybind11/how-to.rst
```diff
@@ -384,12 +398,48 @@ The example contains a custom exception (``ExampleError``) added by the ``LSST_E
...
+
+.. note::
+
+    We attach the ``enum`` values to the class (by passing the ``py::class_`` object ``clsExampleOne`` as the first argument)
+
+.. note::
+
+    Add ``.export_values()`` if (and only if) you need to export the values into the class scope (so they can be reached as ``ExampleOne.RED``, in addition to ``ExampleOne.State.RED``).
+
+    Never do this for new style scoped ``enum class`` types, since that will give them different symantics in C++ and Python.
```
ktlim: I know this was in the original, but maybe "C++11" instead of "new style".

pybind11/how-to.rst
```diff
@@ -401,57 +451,64 @@ The end result of all the steps above looks like this:
...
     #include "pybind11/stl.h"
 
-    #include "numpy/arrayobject.h"
-    #include "numpy/arrayobject.h"
     #include "ndarray/pybind11.h"
 
     #include "lsst/pex/exceptions/python/Exception.h"
 
-    #include "lsst/TMPL/ExampleOne.h"
+    #include "lsst/tmpl/ExampleOne.h"
```
ktlim: "TMPL" is still used above on at least lines 177 and 184; this should be consistent.

pybind11/how-to.rst
```diff
@@ -468,14 +525,15 @@ Edit ``python/.../SConscript`` to look like this:
...
-            'exampleOne',
+            '_tmpl',
             # ... list additional pybind11 wrappers, if any
         ],
         addUnderscore=False,
     )
 
 ``addUnderscore`` is historical baggage; eventually we plan to make ``False`` the default.
 
+
```
ktlim: I guess updating sconsUtils at this time isn't worth it?

pybind11/how-to.rst
```diff
@@ -553,34 +611,86 @@ and ``ExampleThree.h``:
...
 ^^^^^^^^^^^^^^^^^^^^
 
-Again following :ref:`our rules on file naming <style-guide-pybind11-module-naming>` we create a basic file for the wrapper ``python/lsst/TMPL/exampleTwo.cc`` (note that this file will later move to a subpackage):
+Classes defined in different headers should usually be wrapped in different source files, which are then linked into a single compiled module (it may be better to wrap very similar classes in a single source file to make it easier to share wrapper code between them).
+
+.. note::
+
+    An older version of this tutorial advocated compiling each source file as a separate module, which makes it impossible to correctly handle circular dependencies, and generally leads to larger-than-necessary binary module sizes.
+
+A compile module should have exactly one ``PYBIND11_MODULE`` block that creates and finishes a ``WrapperCollection``.
```
ktlim: "compile" -> "compiled"

pybind11/how-to.rst
```diff
@@ -650,105 +771,40 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
-
-Adding these import statements in the C++ wrapper, rather than relying on import order in ``__init__`` in Python, follows :ref:`our rule on import <style-guide-pybind11-import>`.
 
-Finished wrappers (C++ part)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Cross-module dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The end results for the C++ part of the wrappers (see next for the Python part) are ``exampleTwo.cc``:
+All of the depenxencies in the example classes we've defined here are within the same compiled module, and hence we've been able to rely on ``WrapperCollection`` and its callback system to ensure the wrappers are defined in an order that works.
```
ktlim: "depenxencies" -> "dependencies"

pybind11/how-to.rst
```diff
@@ -650,105 +771,40 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
+As its name suggests, the former is needed when you want to inherit from a base class defined in another module; this must be called before the ``py::class_`` instantiation of the derived class.
+``addSignatureDependency`` declares that the external type is used only in function or method signatures.
+Regardless of when it is called, these dependencies will be imported after all local types are declared and before any definition callbacks are run.
 
-    namespace py = pybind11;
-    using namespace pybind11::literals;
-
-    namespace lsst {
-    namespace tmpl {
+Circular *inheritance* dependencies are impossible in both C++ and Python, but circular signatures dependencies are relatively common in C++ (via forward declarations) and quite possible in Python (via function-level imports and duck-typing).
```
ktlim: "signatures" -> "signature"

pybind11/how-to.rst
```diff
@@ -650,105 +771,40 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
+In pybind11 wrappers, circular signature dependencies are a problem.
+Using ``WrapperCollection`` solves that problem within a single module (in which all wrappers are added to the same ``WrapperCollection`` instance), and it takes steps to solve across multiple modules, by executing signature-dependency imports between type definitions and function definitions.
```
ktlim: "takes steps to solve" -> "partially solves this"?

pybind11/how-to.rst
```diff
@@ -650,105 +771,40 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
+But this only solves cross-module circular dependencies to the extent that a circular Python imports result let the second importer see the types in the partially-defined first module.
```
ktlim: This sentence needs much clarification. 😄

pybind11/how-to.rst
```diff
@@ -650,105 +771,40 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
+It's not clearly that something guaranteed by the CPython interpreter (let alone the Python language), so it's much safer too just make sure any types with a circular dependency are handled by the same ``WrapperCollection`` instance and defined in the same module.
```
ktlim: Remove "that".  "too" -> "to".  Is it always possible to have circularly-dependent types defined in the same module?

pybind11/how-to.rst
```diff
@@ -650,105 +771,40 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
-The pure Python code then goes into ``exampleTwo/exampleTwoContinued.py``.
-We shall use the ``continueClass`` decorator to reopen the class and add a new method:
+We'll use the ``continueClass`` decorator to reopen the class and add a new method:
 
 .. code-block:: python
 
     from lsst.utils import continueClass
 
-    from .exampleTwo import ExampleTwo
+    from ._tmpl import ExampleTwo
```
ktlim: Just checking: this should be `._tmpl` and not just `.`?

pybind11/style.rst
```diff
@@ -59,33 +58,40 @@ Modules and source files
 
 .. _style-guide-pybind11-module-naming:
 
-Wrappers for a C++ header file SHOULD go in a Python module with a lowercased version of the header file name
--------------------------------------------------------------------------------------------------------------
+Compiled for a package SHOULD go in a module and source file named by adding an underscore prefix to the package name.
```
ktlim: I think you want a subsection title before this.  "Compiled" doesn't make sense.  Maybe "Top-level wrapper code"?

pybind11/style.rst
```diff
@@ -59,33 +58,40 @@ Modules and source files
...
 
-For example, C++ code from ``LinearTransform.h`` would be wrapped in a module named ``linearTransform``.  If the wrappers are defined purely in C++, the source code would go in ``linearTransform.cc`` (see :ref:`the following rule <style-guide-pybind11-subpackage>` for the case where both C++ and Python code are present).
+For example, the Python module for ``lsst.geom`` should be ``_geom``, defined (at least in part) in a source file `_geom.cc`.
```
ktlim: "the Python wrapper module" instead of just "the Python module"?

pybind11/style.rst
```diff
@@ -59,33 +58,40 @@ Modules and source files
...
 
-By wrapping different headers into separate modules (to be combined in ``__init__``) we make builds more parallelize able, make it easier to avoid circular dependencies, and make partial rebuilds faster.
+.. _style-guide-pybind11-cxx-source:
 
-If a group of headers together provide functionality that cannot be used independently, they may be wrapped into a single module.  The headers wrapped by such a module must be prominently listed in a comment near the top of the source file.
+Wrappers for a C++ header file SHOULD go in a source file with same name as the header and a leading underscore
```
ktlim: "the" before "same name"

pybind11/style.rst
```diff
@@ -59,33 +58,40 @@ Modules and source files
...
+---------------------------------------------------------------------------------------------------------------
+
+For example, C++ code from ``LinearTransform.h`` should be in a source file ``_LinearTransform.cc``.
```
ktlim: "wrapping" instead of "from"

pybind11/style.rst
```diff
@@ -59,33 +58,40 @@ Modules and source files
...
 
-    linearTransform/linearTransformContinued.py:
+    _LinearTransform.py:
         <Python extensions to the wrappers>
 
-    linearTransform/__init__.py:
-        from .linearTransform import *
-        from .linearTransformContinued import *
+    _geom.cc:
+        <C++ module definition>
```
ktlim: Say here that this calls into the code in `_LinearTransform.cc`.

pybind11/style.rst
```diff
@@ -270,13 +254,21 @@ This is especially again common in :ref:`declare functions <style-guide-pybind11
 Organization
 ============
 
+.. _style-guide-pybind11-wrap-split-modules:
+
+Wrappers for a header SHALL be implemented in functions prefixed with "wrap"
+----------------------------------------------------------------------------
+
+Wrappers for ``LinearTransform`` (declared in ``LinearTransform.h``, with wrappers in ``_LinearTransform.cc``) should go in a function called ``wrapLinearTransform``.
```
ktlim: "SHALL" and then "should"?  Sounds wishy-washy.  Also "Typically" below — what are the exceptions?

pybind11/style.rst
```diff
@@ -468,8 +433,7 @@ Keyword arguments MAY be provided for non-overloaded functions with two or fewer
 Literals MUST be used for all named arguments
 ----------------------------------------------
 
-The `_a` argument literal, from `pybind11::literals` MUST be used
-for all named arguments (e.g. ``mod.def("f", f, "arg1"_a, "arg2"_a);``).
+The `_a` argument literal, from `pybind11::literals` MUST be used for all named arguments (e.g. ``mod.def("f", f, "arg1"_a, "arg2"_a);``).
```
ktlim: Remove first comma.

pybind11/how-to.rst
kfindeisen: Please don't defend bad design using words like "common"! People might take it as license to create more circular dependencies. :scream:

pybind11/how-to.rst
TallJimbo: I haven't thought of any, but I dislike being overly prescriptive in a how-to guide, especially having only wrapped one package with this so far.

pybind11/how-to.rst
TallJimbo: I'm not quite sure what your alternate proposal is.  When there are multiple per-header .cc files I don't think it's appropriate to just pick one to drop `PYBIND11_MODULE` in.  But I also don't want to mandate that tiny packages with just one header (e.g. `meas_extension_photometryKron`) have an extra package-level .cc file with just the `PYBIND11_MODULE`.  The problem with this example, as I see it, is that it starts out being a tutorial for the latter case (to start out simple) and then turns into a tutorial for the former case.

pybind11/how-to.rst
TallJimbo: I haven't thought about this in the abstract much, but I don't actually think of circular dependencies *within one library* to be bad design.  In fact, we don't even usually call it that in C++; it's just related classes interoperating with each other.  IMO the fundamental problem here (well, in afw in particular) is that we have been unclear about what "one library" means, and in particular defined it one way (all of afw) in C++ and a different way (subpackages are supposed to be distinct) in Python.

pybind11/how-to.rst
TallJimbo: Maybe I should just delete the second half of this sentence entirely, at least for now; I'm sort of speculating about what Python does when it sees circular imports, and hence I'm definitely speculating on whether `WrapperCollection` will actually make that work.

pybind11/how-to.rst
TallJimbo: Maybe not, but I think the cases where it would be impossible are sufficiently exotic that they're unlikely to ever arise in our system.  I think you'd need to have two packages with their own C++ libraries with one optional package-level dependency and one required one, and the required package would need to have Python bindings that don't depend on any of its own symbols that are provided by the optional library.

pybind11/how-to.rst
TallJimbo: Yes.  Both *may* work, but just `.` relies on `__init__.py` being re-imported while it is already being imported and that partially-imported state providing the `ExampleTwo` symbol.

pybind11/style.rst
TallJimbo: Agree the language is inconsistent, though I might prefer to downgrade the `SHALL` to `SHOULD` - for both that and "typically" below, I feel like we shouldn't make a not-very-important naming rule a very strict one unless we're quite certain there's no valid exception we should defer to developer judgement.  I know we have a general statement about that in style guide introduction, but I don't think that's a substitute for leaving room for and emphasizing the importance of developer judgement throughout.

I especially don't want to wall ourselves in too much until we've tried to apply these guidelines to afw.

pybind11/how-to.rst
TallJimbo: I don't really have any good criteria, but I can say that I don't know how to reset `__module__` for all function wrappers automatically without adding a ton of boilerplate.  To be honest, it's probably just something that you don't realize you need until it bites you, and I'm not really sure what to do about that.

pybind11/how-to.rst
TallJimbo: I'm not sure we've stopped using it everywhere.  These changes also make using the current signature of this sconsUtils  signature awkward in other ways - now there's usually only a single module, and that means we use the `extraSrc` option a lot more.  It probably makes the most sense to fix both of those at the same time once we're sure the old approaches are no longer being used (leaving aside the question of when that might be).

pybind11/how-to.rst
kfindeisen: Within a _module_, maybe, but even then it can make code hard to understand and harder to modify. Besides, IIRC a lot of the circular dependencies in `afw` are two-way type conversions, which could have been handled in non-circular ways, or classes that depend on their subclasses, which really do cause problems (in particular, it's hard to extend the existing code).

In any case, I'm not asking you to forbid circular dependencies, just not encourage them (e.g., replacing "common" -> "possible").

pybind11/how-to.rst
ktlim: I was indeed suggesting that every package have a `_package.cc` file, even if they had only one include file, to make things consistent.  It depends on what you see as the usual case.  If most packages have multiple includes, then I'd say start by describing that case and treat the only-one-include as a special optimization.  If you think most packages will have only one include, then stick with this but make clear that it is only applicable in that case.

pybind11/how-to.rst
TallJimbo: Yes, both still work, but the former is can make pickle files that rely on the function unnecessarily fragile to later refactoring.  I'll rewrite to clearly explain the pickle example, as it's the only one I've actually got.

pybind11/how-to.rst
TallJimbo: I've added an italicized "*within a library*".

pybind11/how-to.rst
TallJimbo: I've reworded this to just say that `WrapperCollection` does its best to solve circular dependencies across modules, but that this relies on how Python itself handles circular dependencies and there are no guarantees.

pybind11/how-to.rst
```diff
@@ -645,110 +781,151 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
+The module names and file/directory structure can be quite confusing in this case, so we'll look at a very concrete example.
+Let's imagine we have a package ``lsst.foo``, with a normal package-level
+module ``_foo``.
+That implies we have the following files:
+::
+    lsst/
+        foo/
+            __init__.py
+            _foo.cc
+            SConscript
```
TallJimbo: @jonathansick, sphinx is warning about "unexpected indentation" in this code block, possibly because it thinks this is Python code.  What should I be doing instead?

(note that it does actually render this correctly, and it did not when I was trying to use `.. code-block::` for it).

pybind11/how-to.rst
```diff
@@ -553,88 +631,146 @@ and ``ExampleThree.h``:
...
+
+    // _tmpl.cc
+
+    #include "pybind11/pybind11.h"
+
+    #include "lsst/utils/python.h"
+
+    namespace lsst { namespace tmpl {
+
+    void wrapExampleOne(utils::python::WrapperCollection & wrappers);
```
ktlim: Maybe it's making things too magic, but I was wondering if you couldn't reduce boilerplate by defining a few preprocessor macros to make this:
```
HEADER(ExampleOne);
HEADER(ExampleTwo);
HEADER(ExampleThree);

PYBIND11_MODULE(_tmpl, mod) {
    WRAP_COLLECTION(mod, "lsst.tmpl"); // automatically gets list of headers from HEADER macros and includes finish
}
```

pybind11/how-to.rst
```diff
@@ -292,32 +313,37 @@ Wrapping a member function is easy:
...
+Following the :ref:`rule on STL containers <style-guide-pybind11-stl-containers>` we simply include the ``pybind11/stl.h`` header (to enable automatic conversion to and from Python containers) and wrap the function as normal:
 
 .. code-block:: cpp
 
-    clsExampleOne.def("computeSomeVector", &ExampleOne::computeSomeVector);
+    cls.def("computeSomeVector", &ExampleOne::computeSomeVector);
+
+.. note::
+
+    The converters defined in ``pybind11/stl.h`` do a complete conversion
```
ktlim: Combine into one line.

pybind11/how-to.rst
TallJimbo: My first thought is that it's too magical, though I could be persuaded.  This would probably work for most packages, but it might be confusing that the order of `HEADER`s can matter (for inheritance dependencies) if it sets the order in `WRAP_COLLECTION`.  It would not work for afw, where we do need binary submodules - it's unfortunately both a common example and a bad one for future code.

FWIW, I don't immediately see how you'd implement `WRAP_COLLECTION`'s getting the list from the `HEADER` invocations - maybe kind of static vector of function pointers?

pybind11/how-to.rst
```diff
@@ -645,110 +781,151 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
-Adding these import statements in the C++ wrapper, rather than relying on import order in ``__init__`` in Python, follows :ref:`our rule on import <style-guide-pybind11-import>`.
+Circular *inheritance* dependencies are impossible in both C++ and Python, but circular signature dependencies are relatively common *within a single library* in C++ (via forward declarations) and quite possible in Python (via function-level imports and duck-typing).
+In pybind11 wrappers, circular signature dependencies are more of a problem than they are in either language independently.
+Using ``WrapperCollection`` solves that problem within a single module (in which all wrappers are added to the same ``WrapperCollection`` instance).
+It also does its best to work around circular dependencies between different modules, but this relies on the details of how Python handles circular imports, making it very hard to guarantee correct behavior in all cases.
+Circular dependencies between modules should be avoided whenever possible.
 
-Finished wrappers (C++ part)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Binary Submodules
```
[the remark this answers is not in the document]
TallJimbo: Would "Binary Subpackages" be better?

pybind11/how-to.rst
ktlim: Yes, let's leave it as is for now.  (I _was_ thinking of a function pointer vector, although namespacing and initialization of that vector were going to be potential issues.)

pybind11/how-to.rst
```diff
@@ -645,110 +781,151 @@ Following :ref:`this rule <style-guide-pybind11-declare-template-wrappers>` we d
...
-    namespace tmpl {
-    namespace {
+    } // namespace bar
 
-    template <typename T>
-    static void declareExampleThree(py::module & mod, std::string const & suffix) {
-        using Class = ExampleThree<T>;
-        using PyClass = py::class_<Class, std::shared_ptr<Class>, ExampleBase>;
+    PYBIND11_MODULE(_foo, mod) {
+        WrapperCollection wrappers;
```
ktlim: Doesn't this need `(mod, "lsst.foo")` arguments?

PR #264, opened 2018-09-25
DM-15858: copyright.py breaks on Python 3.6+

kfindeisen:
This PR fixes a bug where the `git log` output was being read as raw bytes rather than text, breaking parsing.

### DIFF
--- a/legal/copyright.py
+++ b/legal/copyright.py
@@ -171,7 +171,7 @@ if __name__ == "__main__":
     if len(sys.argv) > 1:
         git_cmd += sys.argv[1:]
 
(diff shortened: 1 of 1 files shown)


kfindeisen:
Hmm... the build at https://developer.lsst.io/v/DM-15858/legal/copyright-overview.html#background shows the old file. :disappointed:

jonathansick:
That's an interesting bug. I'll look into it.

jonathansick:
It seems that the upload failed in Travis. Probably a transient error. A rebuild will probably fix it.

kfindeisen:
It works now, thanks!

PR #268, opened 2018-10-04
DM-15973: Fix so package docs don't have modules section

jonathansick:
This is some text unresolved from DM-15024 when we dropped package docs if there was a Python/C++ module present.

There is a new URL for the example file as well (post DM-15024 fix).


kfindeisen:
Actually, I do have one request: can you replace mention of "module" with "package" throughout the file?

jonathansick:
> can you replace mention of "module" with "package" throughout the file?

@kfindeisen Sorry, which file?

kfindeisen:
`stack/package-homepage-topic-type.rst`

jonathansick:
Thanks. I've fixed the issues with both the future components and "modules."

PR #271, opened 2018-10-05
DM-16031: Fix syntax errors and instate build feedback

jonathansick:
This fixes some syntax issues that have appeared in the storage page and in the Pybind11 page.

It also makes the build more stringent against syntax errors and deployment failures so we don't have these issues in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,8 @@ install:
   - pip install -r requirements.txt
   - pip install "ltd-mason>=0.2,<0.3"
 script:
-  # TODO Add -W flag (DM-9815)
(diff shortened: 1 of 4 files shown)


jonathansick:
@TallJimbo can you verify that the fixes to [pybind11/how-to](
https://developer.lsst.io/v/DM-16031/pybind11/how-to.html) are ok?

@mbutlerncsa can you verify that the fixes to the [Storage doc](
https://developer.lsst.io/v/DM-16031/services/storage.html) are ok?

jonathansick:
Good catch.

PR #278, opened 2018-10-23
Made section more generic.

gcomoretto:


jdswinbank:
Is the intention of this to exclude the possibility of third party code for everything except Science Pipelines? Doesn't Qserv rely on Boost, for e.g.?

gcomoretto:
From my understanding, the issue is to try to make the text more close to the reality. If in the future other top level packages will use externals, the text need to be rephrased.

jdswinbank:
Qserv is a top-level product which depends on [a bunch](https://github.com/lsst/qserv/blob/master/ups/qserv.table#L1) of third party packages (antlr, Boost, Doxygen, Flask, ...), so it's a clear counter-example.

Regardless, it's not clear whether the text as written is normative (ie, establishing the rule that non-Pipelines code _may not_ depend on externals) or descriptive (stating that, as of now, it _does not_, but leaving that open to change in future).

timj:
I don't understand why this needs to change at all.  Third party code should be in those teams regardless of who is using it.

gcomoretto:
Closed without merging.

PR #279, opened 2018-10-25
DM-16296: Add W504 to style guide

timj:
This matches the style already used in the guide and in our code.

### DIFF
--- a/python/style.rst
+++ b/python/style.rst
@@ -61,6 +61,14 @@ E228
 Maximum line length
    See :ref:`style-guide-py-line-length`.
 
+The style checker in ``pycodestyle`` also provides warnings that can be used to request a specific style that is ambiguous in :pep:`8`.
+These codes should be ignored to choose the LSST preferred style:
+
+W504
+   Line break after binary operator.
+   Disabling this enables W503 that checks that line breaks do not occur before binary operators.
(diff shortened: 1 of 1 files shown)


jdswinbank:
Could we please augment this with a definitive statement about where newlines are supposed to go, that isn't confused by a bunch of tool-specific error codes?

As far as I'm aware (which may not be very far...), there has been no unambiguously documented requirement to date about where to put your newline with respect to your operator. I think that this is effectively introducing one, based on codifying past practice and extrapolating from examples. That's ok, I suppose, but let's make that requirement obvious rather than burying it in a discussion about tooling.

ktlim:
My suggestion is to add an explicit item to section 3 "Whitespace" that says something like:

> Wrap lines after binary operators
> 
> PEP 8 does not prescribe whether lines should be broken before or after binary operators.  For consistency with our C++ code (reference https://developer.lsst.io/cpp/style.html#the-incompleteness-of-split-lines-must-be-made-obvious), we choose after.
> 
> This requires that W504 be disabled in pycodestyle.

I do not think this requires an RFC since it is documenting current practice.

I would welcome an RFC to change both Python and C++ to the Knuth-approved style.

timj:
@ktlim I've added your suggestion to the pull request.

PR #284, opened 2018-11-12
Add log configuration examples

parejkoj:
There were no examples of how to configure subtask logging on the commandline or in a config.

Here's the built version: https://developer.lsst.io/v/u-parejkoj-add-log-config/stack/logging.html#logging-within-the-task-framework


ktlim:
I'm not sure I like setting logging configuration in the Task config.  First, having more general Python in it may cause problems in the long run for freezing and round-tripping.  Second, I think that in the not-too-distant future we will want to separate runtime configuration like logging and Butler settings from algorithmic configuration.  But the command line example is fine.

parejkoj:
I've removed the "configure in config" section, and fixed up the other text a bit.

PR #286, opened 2018-11-20
Describe permitted styles for Python imports.

jdswinbank:
This is intended as a stop-gap until RFC-34 is implemented.

### DIFF
--- a/python/style.rst
+++ b/python/style.rst
@@ -545,6 +545,28 @@ methods, ``cls`` should be used instead.
     This is consistent with the naming conventions in PEP 8 as indicated explicitly
     by `upstream <https://mail.python.org/pipermail/python-dev/2018-January/151986.html>`_.
 
+.. _style-guide-py-naming-import:
+
+Import statements SHOULD follow a standard pattern
+--------------------------------------------------
+
+You may choose either to write module import statments so that they import the fully-qualified module name:
+
+.. code-block:: py
+
+   import lsst.foo.bar
+
+or you may use ``as`` to assign a short name to the module:
+
+.. code-block:: py
+
+   import lsst.foo.bar as fooBar
+
+Short names are typically formmed by dropping ``lsst.`` and combining the rest of the name using camel-case.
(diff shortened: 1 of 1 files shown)


jonathansick:
@jdswinbank I don't get the connection to RFC-45 here?

jdswinbank:
Offset by one key to the left...!

jdswinbank:
Reference discussion at https://lsstc.slack.com/archives/C2JPL8U78/p1542743358082100.

TallJimbo:
It's not clear to me whether this text rules out a third style:
```
from lsst.foo import Bar
```
That's also in common use (and is my current favorite for new code), so I'd at least like to avoid accidentally disallowing it.

python/style.rst
```diff
@@ -545,6 +545,25 @@ methods, ``cls`` should be used instead.
     This is consistent with the naming conventions in PEP 8 as indicated explicitly
     by `upstream <https://mail.python.org/pipermail/python-dev/2018-January/151986.html>`_.
 
+.. _style-guide-py-naming-import:
+
+Import statements SHOULD follow a standard pattern
+--------------------------------------------------
+
+When creating a new file, you may choose either to write import statments so that they import the fully-qualified module name:
```
ktlim: "write import statments" → "write module import statements" (to take into account Jim's request).

ktlim: Also, I'd say "when creating a new package".  Each file in a package should endeavor to be consistent with other files in the same package.

ktlim: And it might be nice to mention that an abbreviation chosen should be the same as for any C++ code in the same package (see C++ rule 5-42).

jdswinbank:
Thanks @ktlim. I tweaked the wording a bit based on your suggestions (I like the idea to be consistent with the package, but some of our packages area already not consistent, so I tried to word around that), and will now merge.

ktlim:
Ah, but "statments" still leaked in 😄

jdswinbank:
Rats! I'd fix, but I see you're way ahead of me.

PR #294, opened 2018-12-20
Remove pyvenv instructions from documenteer installation steps

jonathansick:
The pyvenv strategy for isolating documenteer from the Stack's conda environment seems to be error prone based on user feedback. It's easy to use the wrong Python and end up with a Python environment that doesn't make sense. With this commit we're removing mention of the virtual environment and instead recommending that the documenteer package is installed directly into the stack's conda environment. This emulates the likely endgame anyhow, where documenteer is packaged with conda and distributed as part of the conda environment.

- https://developer.lsst.io/v/DM-16935/stack/building-single-package-docs.html
- https://developer.lsst.io/v/DM-16935/stack/building-pipelines-lsst-io-locally.html

### DIFF
--- a/stack/building-pipelines-lsst-io-locally.rst
+++ b/stack/building-pipelines-lsst-io-locally.rst
@@ -42,7 +42,7 @@ Then set up the `pipelines_lsst_io`_ package with EUPS:
 
 .. warning::
 
-   If you’ve already set up packages, you might need to unset-up them before setting up ``pipelines_lsst_io``.
+   If you’ve already have packages set up with the :command:`setup` command, you might need to un-setup them with the :command:`unsetup` command before running ``setup -r pipelines_lsst_io``.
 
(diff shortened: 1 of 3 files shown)


stack/building-pipelines-lsst-io-locally.rst
```diff
@@ -52,43 +52,25 @@ Install Documenteer, the documentation tooling
...
 .. code-block:: bash
 
-   python -m venv --system-site-packages --without-pip pyvenv
-   source pyvenv/bin/activate
-   curl https://bootstrap.pypa.io/get-pip.py | python
-   pyvenv/bin/pip install -r pipelines_lsst_io/requirements.txt
+   cd pipelinies_lsst_io
+   pip install -r requirements.txt
+
+Do this *after* setting up the EUPS Stack.
```
jdswinbank: (I realise the terminology isn't new here, but...) maybe avoid using “setting up”? It's not immediately obvious whether it means “installing” or running `setup <package>`.

PR #301, opened 2019-01-16
Additional cases when RFC need CCB approval

gcomoretto:


jdswinbank:
- “DRP” doesn't need a definite article (ie, just “generated by DRP”, not “the DRP”).
- @jonathansick would tell you to avoid the “e.g.” — how about “updates to a data model such as the catalog...”.

Otherwise, looks great. Sorry for being slow to review.

PR #302, opened 2019-01-18
Add some text about requirements documents.

ktlim:


project-docs/change-controlled-docs.rst
```diff
@@ -44,6 +44,7 @@ Document formats
 ================
 
 Change controlled documents can be written as Word documents or as LaTeX projects in GitHub repositories.
+Requirements documents are special and are written in MagicDraw.
```
jdswinbank: “Generated from” rather than “written in”?

PR #307, opened 2019-02-08
Mention max-doc-length flake8 configuration

timj:


timj:
I'm having trouble working out phrasing. Are you talking about W504 and max-doc-length as items that are not specified by PEP8 but are enforced by flake8?

ktlim:
Yes.  The note says:

> Flake8 only validates code against PEP 8 specifications.

But this is not exactly true; it also validates some of the LSST-specific variations.  The point that it doesn't validate all of our style guide is of course true and appropriate.

timj:
@ktlim I've tweaked that sentence. Let me know if this looks okay to you.

PR #308, opened 2019-02-13
Describe how to handle class and static methods.

ktlim:


stack/deprecating-interfaces.rst
```diff
@@ -59,6 +72,10 @@ A deprecated pybind11-wrapped function must be rewrapped in pure Python by manua
 If only one overload of a set is being deprecated, state that in the reason string.
 Over-warning is considered better than under-warning in this case.
 
+For wrapped class methods, it is necessary to use more complicated syntax::
+
+    Class.function = classmethod(deprecated(reason="why", category=FutureWarning)(Class.function.__func__))
```
parejkoj: I tried this in `exposureContinued.py` with:

```ExposureF.getCalib = deprecated(reason="Replaced with getPhotoCalib (will be removed in 18.0)",
                                category=FutureWarning)(ExposureF.getCalib.__func__)
```

But that gives me this exception:

```tests/test_exposure.py:36: in <module>
    import lsst.afw.image as afwImage
python/lsst/afw/image/__init__.py:37: in <module>
    from .exposure import *
python/lsst/afw/image/exposure/__init__.py:24: in <module>
    from .exposureContinued import *
python/lsst/afw/image/exposure/exposureContinued.py:104: in <module>
    category=FutureWarning)(ExposureF.getCalib.__func__)
../../lsstsw/miniconda/envs/lsst-scipipe/lib/python3.6/site-packages/deprecated/classic.py:238: in deprecated
    raise TypeError(repr(type(wrapped)))
E   TypeError: <class 'builtin_function_or_method'>
```

Trying `Exposure.getCalib` instead results in `E   AttributeError: type object 'Exposure' has no attribute 'getCalib'`, which is probably true (since the `ExposureT` types are the real ones).

parejkoj: @jdswinbank helped me figure out a viable approach. I think the problem with your above version is that `ExposureF.getCalib` is an `<instanceMethod>`, not a `<function>`, so the `deprecated` code [on line 225 here](https://github.com/tantale/deprecated/blob/master/deprecated/classic.py#L225) will register False and it will fall into the `raise`. I'm not sure if there's a `pybind11` or `deprecated` patch (for the latter, maybe PR that `elif` to read `elif inspect.isfunction(wrapped) or inspect.ismethod(wrapped)`) to fix that interaction?

```python
    @deprecated(reason="Replaced with getPhotoCalib (will be removed in 18.0)", category=FutureWarning)
    def getCalib(self, *args, **kwargs):
        return self._getCalib(*args, **kwargs)
```

where the pybind11 looks like: `cls.def("_getCalib", &ExposureT::getCalib);`

parejkoj:
I've filed an Issue with `deprecated` about the question of depreaction-wrapping pybind11 methods: https://github.com/tantale/deprecated/issues/6

stack/deprecating-interfaces.rst
ktlim: I think patching `deprecated.sphinx` at line 96 with the following code will take care of this in a relatively clean way, doing the equivalent of the above but without requiring changing our code:
```python
        try:
            wrapped.__doc__ = docstring
        except AttributeError:
            def internal(*args, **kwargs):
                return wrapped(args, kwargs)
            internal.__doc__ = docstring
            wrapped = internal
```

parejkoj:
Now that you've closed DM-19290, can you please update this text to reflect that usage?

PR #310, opened 2019-02-20
Added Board Recommended status details

gcomoretto:
I have added some text in the Exceptions and appeals section, trying to capture what was discussed today in the DMCCB#6 and in DMCCB#3.

I did not mention that the workflow needs to be updated.

### DIFF
--- a/communications/rfc.rst
+++ b/communications/rfc.rst
@@ -177,6 +177,12 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
 
 In all of these cases, the RFC is marked as **Flagged** to indicate that it is being considered by the DMCCB.
 
+As a result of the discussion on a **Flagged** RFC, the DMCCB may decide to approve it and move forward with the proposal.
+In this case, the RFC status is set to **Board Recommended**.
+Once one or more triggering issues have been opened on the RFC, the assignee has to change the status to **Adopted**.
+
+All RFCs on DM controlled documents will set to **Adopted** directly by the DMCCB, since they do not require implementation issues.
+
 .. _rfc-formalities:
 
 Formalities
--- a/communications/rfc.rst
+++ b/communications/rfc.rst
@@ -179,9 +179,12 @@ In all of these cases, the RFC is marked as **Flagged** to indicate that it is b
 
 As a result of the discussion on a **Flagged** RFC, the DMCCB may decide to approve it and move forward with the proposal.
 In this case, the RFC status is set to **Board Recommended**.
-Once one or more triggering issues have been opened on the RFC, the assignee has to change the status to **Adopted**.
+The assignee can then attach the triggering issues and adopt the RFC following the standard process.
 
-All RFCs on DM controlled documents will set to **Adopted** directly by the DMCCB, since they do not require implementation issues.
+If the DMCCB does not approve the RFC, its status will be set to **Withdrawn**.
+If the DMCCB declines to consider the RFC, the DMCCB will change the status back to **Proposed**.
+
+All RFCs on DM controlled documents will be set directly to **Adopted** by the DMCCB, since they do not require implementation issues.
 
 .. _rfc-formalities:
 


communications/rfc.rst
```diff
@@ -177,6 +177,12 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
 
 In all of these cases, the RFC is marked as **Flagged** to indicate that it is being considered by the DMCCB.
 
+As a result of the discussion on a **Flagged** RFC, the DMCCB may decide to approve it and move forward with the proposal.
+In this case, the RFC status is set to **Board Recommended**.
+Once one or more triggering issues have been opened on the RFC, the assignee has to change the status to **Adopted**.
+
+All RFCs on DM controlled documents will set to **Adopted** directly by the DMCCB, since they do not require implementation issues.
```
timj: Consider changing the text a little to be something like:

> ...documents will be set directly to **Adopted** by the DMCCB,...

communications/rfc.rst
```diff
@@ -177,6 +177,12 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
 
 In all of these cases, the RFC is marked as **Flagged** to indicate that it is being considered by the DMCCB.
 
+As a result of the discussion on a **Flagged** RFC, the DMCCB may decide to approve it and move forward with the proposal.
```
timj: The "may" here implies that we should say what happens if DMCCB do not approve.

communications/rfc.rst
```diff
@@ -177,6 +177,12 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
...
+In this case, the RFC status is set to **Board Recommended**.
+Once one or more triggering issues have been opened on the RFC, the assignee has to change the status to **Adopted**.
```
timj: Maybe instead say:

> The assignee can then attach the triggering issues and adopt the RFC following the standard process.

?

communications/rfc.rst
gcomoretto: Fine, I add some text about what happens if the DMCCB do not approve the change.

communications/rfc.rst
gcomoretto: I am not convinced about the use of the verb "can".

timj: Do you prefer `should`?  I was trying not to imply we were forcing people to adopt the RFC.

gcomoretto: OK, It can be used "can" as you suggest, and add at the end ".... standard process, or reopen it setting its status back to **Flagged**."

communications/rfc.rst
```diff
@@ -177,6 +177,15 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
 
 In all of these cases, the RFC is marked as **Flagged** to indicate that it is being considered by the DMCCB.
 
+As a result of the discussion on a **Flagged** RFC, the DMCCB may decide to approve it and move forward with the proposal.
+In this case, the RFC status is set to **Board Recommended**.
+The assignee can then attach the triggering issues and adopt the RFC following the standard process or reopen it changing its status to **Fagged**.
```
timj: Fagged -> Flagged

I'm not sure we need to say that last bit though. "following the standard process" is sufficient.

communications/rfc.rst
```diff
@@ -177,6 +177,15 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
...
+
+In the case that the DMCCB is not approving the RFC, its status will be set to **Withdrawn**.
```
timj: "If the DMCCB does not approve the RFC, its status will be set to **Withdrawn**."

I don't think we discussed this at the meeting but I guess that's the reality.

communications/rfc.rst
```diff
@@ -177,6 +177,15 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
...
+Tn the case that the DMCCB consider that the RFC do not require its approval, the RFC status will be set back to **Proposed**.
```
timj: "If the DMCCB declines to consider the RFC, the DMCCB will change the status back to **Proposed**."

communications/rfc.rst
```diff
@@ -177,6 +177,15 @@ If an RFC includes API changes, including deprecations, or updates to a data mod
...
+
+All RFCs on DM controlled documents will set directly to **Adopted** by the DMCCB, since they do not require implementation issues.
```
timj: "will be set"

communications/rfc.rst
jdswinbank: So you two are the DMCCB authorities, but:

Should the DMCCB not be considering the RFC *after* the triggered tickets have been created, rather than before?

We've been strict that RFCs may not be adopted before the triggered tickets are created, as there's too much scope for misreading a lengthy discussion or believing consensus has been reached when it hasn't really. Making the triggered tickets is supposed to make the concrete outcome of the RFC clear.

It seems to me that it's that concrete outcome which the DMCCB should be considering.

(And yes, I realise that isn't what was discussed at the meeting earlier; I'm just trying to think through the implications.)

timj: That's an interesting idea. Insisting on the triggered tickets being there before CCB weigh in is worth considering. It would mean that CCB would be adopting the RFC directly rather than the reporter.

gcomoretto: I was just trying to capture the conclusions we reach during the CCB telecons.

@jdswinbank , it seems that you have doubts about the process we are following. Shall I cancel or suspend this PR and bring the topic again for discussion in the next DMCCB?

jdswinbank: So I have a few different thoughts in response to that —

- As Tim says, I thought it was an interesting idea, and worth suggesting. I didn't intend it to be a demand, or even as strong as a request. However, if you reckon it's worthy of further discussion, I'd be happy to participate in that.
- That said, I will be travelling 2 through 24 March, and won't be at the next DMCCB meeting to discuss it.
- In general (for this and other policy or procedural discussions), my suggestion would be that the best thing to do is for one individual to write out a detailed proposal, thinking through all of the ramifications, before proposing it to the meeting for approval. I'm concerned that *any* proposal we come up with at the tail end of a 30 minute timeboxed DMCCB telecon will be inadequately thought through and leave a bunch of loose ends hanging.

gcomoretto: I suggest we implement in the dev guide the changes agreed during the DMCCB #3 and #6, and we defer further discussions.

It can be just an AOB point in the next DMCCB, or, if the discussion may be too long, become an RFC.

timj:
We discussed keeping "Board Recommends" (since how else do we indicate to the proposer that everything is fine with us) but changing the next state transition so it is no long "Manager Approves" but is "Adopt" as for anything else.

PR #312, opened 2019-02-26
DM-11462: Start the user documentation style guide

jonathansick:
This PR implements the idea of a user documentation content style guide that's based on Google's, with tweaks for LSST use cases. [RFC-555](https://jira.lsstcorp.org/browse/RFC-555)

There are a lot of additional areas this can go. I'd like to compile a bibliography of technical communication books, blogs, and talks, for instance. But this scope is enough for the style guide project to start being useful.

Preview of the entrypoint to the guide: https://developer.lsst.io/v/DM-11462/index.html#part-user-doc-style-guide

### DIFF
--- a/user-docs/lsst-specific-content-style-guide.rst
+++ b/user-docs/lsst-specific-content-style-guide.rst
@@ -1,0 +1,33 @@
+.. _user-doc-style-lsst:
+
+#########################
+LSST-specific style guide
+#########################
+
+The :doc:`LSST user documentation style guide </user-docs/index>` is built upon the `Google Developer Style Guide`_.
+This page lists our exceptions and modifications to those guidelines.
+
+Use the imperative mood for Python and C++ function and method summary sentences
+--------------------------------------------------------------------------------
+
+For LSST DM Python and C++, you should use the **imperative mood** to write the summary sentence:
+
+   Get the value.
+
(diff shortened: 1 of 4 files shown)


afausti:
Looks great! The page has information and links to user documentation style guide in the DM developer guide. It clarifies where to use the style guide and contexts where the style guide does not apply.

user-docs/lsst-specific-content-style-guide.rst
```diff
@@ -0,0 +1,31 @@
...
+--------------------------------------------------------------------------------
+
+The Google Developer Style Guide recommends `using the present tense for function and method summaries <https://developers.google.com/style/api-reference-comments#methods>`__.
+For example:
+
+   Gets the value.
+
+For LSST DM Python and C++, you should instead use the **imperative mood** in this context:
+
+   Get the value.
```
parejkoj: I would suggest reversing the order of these examples:

```
For LSST DM Python and C++, use imperative:
  ...
but if you are documenting an unrelated technology, for example HTTP API endpoints, use present tense, as recommended by the Google guide:
  ...
(discussion of rationale)
```

That way, the first thing a style guide user sees is the "correct" LSST usage for most cases.

jonathansick: Good idea

PR #321, opened 2019-05-08
DM-19681: Clarify who has authority to override coding standards

jdswinbank:


ktlim:
The problem with this rewrite is that there is then no effective difference between "REQUIRED" and "MUST/SHALL".    Perhaps we want this, but then it should be made more clear by grouping those two items together.

jdswinbank:
> The problem with this rewrite is that there is then no effective difference between "REQUIRED" and "MUST/SHALL". 

I'm not sure that's true.

There's certainly no difference in procedure as regards asking for exceptions, but there's a more ominous warning: by my reading, a “MUST/SHALL” is something the Architect is quite likely to be flexible about if you ask nicely, but a "REQUIRED" will require much greater justification.

I think it's your call as to whether you regard that distinction as worthwhile (you are, after all, making the decision in either case), but unless you would like to propose alternative wording now, I suggest that it makes sense to merge this as is — it's clearly better to refer to people that actually exist. :-)

PR #322, opened 2019-05-29
Updates to DRP computing details

PaulPrice:


PaulPrice:
Branch build at: https://developer.lsst.io/v/u-price-20190529/team/drp.html

team/drp.rst
```diff
@@ -68,12 +62,17 @@ To initialize the stack in your ``bash`` shell, run:
 .. _Tiger: http://www.princeton.edu/researchcomputing/computational-hardware/tiger
 .. _Perseus: http://www.princeton.edu/researchcomputing/computational-hardware/perseus
 
+The Princeton astronomical software group owns a head node on the Tiger cluster called ``tiger2-sumire``.
+You can use this node for building software and running small and/or short-lived jobs.
+
+
```
jdswinbank: Suggest using only one blank line for consistency with the rest of this file.

team/drp.rst
```diff
@@ -68,12 +62,17 @@ To initialize the stack in your ``bash`` shell, run:
...
 Storage
 -------
 
 HSC data (both public data releases and private data, which may not be shared outside the collaboration) is available in :file:`/tigress/HSC` on both clusters.
 This filesystem is available from both clusters, and you may use it to store your results.
 However, note that space is at a premium, especially during our periodic HSC data release processing: please clean up any data you are not actively using.
+Also, be sure to set :command:`umask 002`.
```
jdswinbank: Maybe add a word of explanation as to why.

team/drp.rst
```diff
@@ -98,14 +97,31 @@ Options include:
 
 - Bouncing your connection through a `host on the Peyton network <http://www.astro.princeton.edu/docs/Hardware>`_ (this is usually the easiest way to go);
 - Making use of the `University's VPN service <https://www.net.princeton.edu/vpn/>`_.
+- Using the Research Computing gateway.
```
jdswinbank: Given the Research Computing gateway, is there still a reason to bounce through Peyton? If not, it might be an idea to remove that suggestion for simplicity's sake.

team/drp.rst
```diff
@@ -98,14 +97,31 @@ Options include:
...
+
+(It may also be necessary to add a ``User`` line under ``Host tigressgateway`` if there is a mismatch between your local and Princeton usernames.)
+Entry to ``tigressgateway`` requires `2FA <https://www.princeton.edu/duoportal>`_;
+we recommend using the ``ControlMaster`` feature of SSH to persist connections, e.g.::
+
+    ControlMaster auto
+    ControlPath ~/.ssh/controlmaster-%r@%h:%p
+    ControlPersist 30d
+
+
```
jdswinbank: Again, one blank.

team/drp.rst
```diff
@@ -98,14 +97,31 @@ Options include:
...
+    Host tiger* perseus* tigressdata*
+        ProxyCommand ssh -q -W %h:%p tigressgateway.princeton.edu
+
+(It may also be necessary to add a ``User`` line under ``Host tigressgateway`` if there is a mismatch between your local and Princeton usernames.)
+Entry to ``tigressgateway`` requires `2FA <https://www.princeton.edu/duoportal>`_;
+we recommend using the ``ControlMaster`` feature of SSH to persist connections, e.g.::
+
+    ControlMaster auto
+    ControlPath ~/.ssh/controlmaster-%r@%h:%p
+    ControlPersist 30d
```
jdswinbank: I do this (except I persist for 60s, not 30 days!), but I find it quite often causes problems — network glitches can cause the master connection to drop, but SSH doesn't always realise it and tries to reuse it for other sessions. Nothing works until I `rm ~/.ssh/master-*`.

This may be some weird peculiarity of my system (I've never tried hard to debug), but if it happens to others recommending this approach may cause more confusion than it resolves.

team/drp.rst
PaulPrice: I left it in, as it may still be the easiest way to go (no 2FA?).

team/drp.rst
PaulPrice: Reduced to `10m`.

PR #327, opened 2019-06-25
Add required teams to repo creation.

ktlim:


ktlim:
Doesn't seem like there's a trivial way to check this in `lsst/repos:bin/repos-lint`, as I think it requires authentication to use the list-teams API.  If there's some easy way to provide GitHub auth to Travis, though, it would be nice to do so.

PR #331, opened 2019-07-16
DM-5184: Note that EUPS wants unified diffs.

jdswinbank:


stack/packaging-third-party-eups-dependencies.rst
```diff
@@ -205,8 +205,7 @@ with, e.g. :file:`000-something.patch`, :file:`001-somethingelse.patch`.
 
 .. note::
 
-   EUPS expects the patches to be formatted according to the output of
-   :command:`git diff`, not the output of :command:`diff`.
+   EUPS expects the patches to be in unified format, as generated by the ``-u`` option to the :command:`diff` command.
```
timj: The output of `git diff` does also work... (but for third party packages you are extremely unlikely to be able to use that command to generate the diff).

jdswinbank: That is because ``git diff`` generates unified diffs. :-)

parejkoj: Maybe note that in a parenthetical? Otherwise, this looks good.

jdswinbank: I kinda feel the parenthetical would be less helpful than it would be confusing... you won't be using git diff unless you aren't following the instructions.

PR #337, opened 2019-07-30
DM-7450: Add note about shebang in test files

timj:
Reopens #121 with the text in a different place.

### DIFF
--- a/python/testing.rst
+++ b/python/testing.rst
@@ -13,7 +13,8 @@ LSST tests should be written using the `unittest` framework, with default test d
 If you want to jump straight to a full example of the standard LSST Python testing boilerplate without reading the background, read the :ref:`section on memory testing <py-test-mem>` later in this document.
 
 .. note::
-   Python tests explicitly should not contain a shebang (``#!/usr/bin/env python``) and should not be executable (run directly with ``./test_Example.py``).
+   Python tests explicitly should not contain a shebang (``#!/usr/bin/env python``) and should not be executable (so cannot be run directly with ``./test_Example.py``).
+   Instead run single test files either with ``python tests/test_Example.py`` or with ``pytest tests/test_Example.py``.
    This avoids problems encountered running tests on macOS and helps ensure consistency in the way that tests are executed.
 
 .. _SQR-012: http://sqr-012.lsst.io
--- a/python/testing.rst
+++ b/python/testing.rst
@@ -303,16 +303,16 @@ Arguments can be provided to the ``args`` keyword parameter in the form of a seq
    When the codebase has migrated to consistently use the testing scheme described in this section :lmod:`~lsst.sconsUtils` will be modified to disable the duplicate testing.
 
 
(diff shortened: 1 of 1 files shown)


python/testing.rst
```diff
@@ -12,6 +12,11 @@ See :doc:`/coding/unit-test-policy` for an overview of LSST Stack testing.
 LSST tests should be written using the `unittest` framework, with default test discovery, and should support being run using the `pytest`_ test runner as well as from the command line.
 If you want to jump straight to a full example of the standard LSST Python testing boilerplate without reading the background, read the :ref:`section on memory testing <py-test-mem>` later in this document.
 
+.. note::
+   Python tests explicitly should not contain a shebang (``#!/usr/bin/env python``) and should not be executable (so cannot be run directly with ``./test_Example.py``).
+   Instead run single test files either with ``python tests/test_Example.py`` or with ``pytest tests/test_Example.py``.
```
jonathansick: Why would one ever run the test file with `python` rather than `pytest`? Oh; because stack tests have an `if __name__ == "___main__" to run `unittest.main()`.

python/testing.rst
```diff
@@ -12,6 +12,11 @@ See :doc:`/coding/unit-test-policy` for an overview of LSST Stack testing.
 LSST tests should be written using the `unittest` framework, with default test discovery, and should support being run using the `pytest`_ test runner as well as from the command line.
 If you want to jump straight to a full example of the standard LSST Python testing boilerplate without reading the background, read the :ref:`section on memory testing <py-test-mem>` later in this document.
 
+.. note::
```
jonathansick: This note seems out of place. It'd try to put this note next to the boilerplate example of a test file. I'm not entirely sure what that would be. Either in "Introduction to unittest" or "Memory and file descriptor leak testing" (which is where the full boilerplate seems to be tucked into).

python/testing.rst
timj: It's *significantly* faster to run up the test if you are iterating quickly on things. Also, subTest works properly so in some cases when you are testing things it is clearer to use subTest to get all the failures from the loop in one go than it is to run pytest and get one failure then rerun and get the next failure in the loop.

python/testing.rst
timj: I'll have a look. Reminds me that the memory test no longer tests memory...

timj:
@jonathansick I've had another go at this so please take another look.

python/testing.rst
```diff
@@ -12,6 +12,8 @@ See :doc:`/coding/unit-test-policy` for an overview of LSST Stack testing.
 LSST tests should be written using the `unittest` framework, with default test discovery, and should support being run using the `pytest`_ test runner as well as from the command line.
 If you want to jump straight to a full example of the standard LSST Python testing boilerplate without reading the background, read the :ref:`section on memory testing <py-test-mem>` later in this document.
 
+
```
jonathansick: trim this whitespace?

PR #352, opened 2019-10-17
DM-20267: Add required README.txt to /datasets/refcats/htm

parejkoj:
While creating the reference catalog readme file, I realized we could use more guidance about how to manage the contents of this file.

### DIFF
--- a/services/datasets.rst
+++ b/services/datasets.rst
@@ -43,7 +43,40 @@ Reference Catalogs
 For the Gen2 Middleware, reference catalogs are contained in the repository itself, in a ``ref_cats/`` subdirectory.
 For ``/datasets`` repositories, we handle this by symlinking from ``repo/ref_cats/NAME`` to the corresponding refcat directory in ``/datasets/refcats``.
 The version subdirectory (e.g. ``v0/``, ``v1``) should match the ``REFCAT_FORMAT_VERSION`` that is set by the refcat ingestion task.
-When adding a refcat, don't forget to add a note to the relevant ``README.txt``, as described in the `responsibilities`_ section.
+
+When adding a refcat, you have a :ref:`responsibility <responsibilities>` to supply a ``README.txt`` for the new refcat, and update the overall ``README.txt`` for that type of refcat (e.g. ``/datasets/ref_cats/htm/README.txt``).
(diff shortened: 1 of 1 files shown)


services/datasets.rst
```diff
@@ -43,7 +43,16 @@ Reference Catalogs
...
 For ``/datasets`` repositories, we handle this by symlinking from ``repo/ref_cats/NAME`` to the corresponding refcat directory in ``/datasets/refcats``.
 The version subdirectory (e.g. ``v0/``, ``v1``) should match the ``REFCAT_FORMAT_VERSION`` that is set by the refcat ingestion task.
-When adding a refcat, don't forget to add a note to the relevant ``README.txt``, as described in the `responsibilities`_ section.
+
+When adding a refcat, don't forget to add a note to the relevant ``README.txt``, as described in the `responsibilities`_ section. Updates to the readme should be reviewed on a Jira ticket about the new reference catalog (for example, `DM-20267 <https://jira.lsstcorp.org/browse/DM-20267>`_). Each entry should have at minimum the following information:
+
+::
+
+    Original data: https://www.example.com/DataRelease9000
+    Jira Epic: https://jira.lsstcorp.org/browse/DM-XXXXX
```
ktlim: Why is this "Epic" rather than "readme review ticket" or "RFC implementation ticket", and why is it needed in the first place?  Is it meant to capture information about why or how the refcat was added?

parejkoj: I wanted to make sure that the tickets where the work was done are linked in the readme. An Epic is about the right amount of work for putting together a refcat: it's certainly several tickets worth. The RFC itself is just signing off on moving the files into place, so it might not capture all the relevant info that went into the construction (though the RFC should link to the relevant Epic).

parejkoj:
I've expanded the readme template and updated the descriptions of what goes where. Please have another look.

Built version: https://developer.lsst.io/v/DM-20267/services/datasets.html#reference-catalogs

PaulPrice:
The "responsibility" link is broken.

services/datasets.rst
PaulPrice: I would have said "ticket" rather than "epic". If it's a full epic, you can put that, but it's not always the case that there's a whole epic devoted to construction of a catalog.

PR #358, opened 2019-11-13
DM-22207: Update documentation to clarify that classes can also be deprecated

leannep:


stack/deprecating-interfaces.rst
```diff
@@ -74,6 +74,9 @@ A deprecated pybind11-wrapped function must be rewrapped in pure Python using th
 If only one overload of a set is being deprecated, state that in the reason string.
 Over-warning is considered better than under-warning in this case.
 
+.. note:
```
leannep: @jonathansick this did not render - how do I add a 'note'

timj: Two colons

PR #359, opened 2019-11-14
DM-22219: Add editor page for VSCode.

TallJimbo:


TallJimbo:
> Possibly make a note about setup environments that people might have trigger from .bash_profile at login.

I've already forgotten exactly what you meant by this, though I imagine we discussed it in person at one point.  I'm going to merge as-is, and let you update this later if you like to avoid a big context switch for either of us.

PR #360, opened 2019-11-18
DM-22199: Add decorators for iterating over tests

PaulPrice:
These make it easy to iterate over parameter combinations.


python/testing.rst
```diff
@@ -342,6 +342,60 @@ final test suite.
...
+            ...
+
+will run tests:
+
+.. code-block:: python
+
+        testSomething(foo=1, bar=3)
+        testSomething(foo=2, bar=4)
+
+Note that the method being decorated must be a subclass of ``unittest.TestCase``, since it relies on the existence of the ``subTest`` method for identifying the individual iterations.
```
ktlim: The method must be _within_ (or on) a subclass of `unittest.TestCase`; the method itself is not a subclass, of course.  You should note that use of `subTest` means that _all_ iterations will be executed, not stopping at the first failure.
You might also "see also" `pytest.mark.parametrize`, which is more complex but offers similar functionality and which we do not yet allow but may soon.

PR #368, opened 2020-01-23
DM-23162: Implement RFC-650 and state that we want to break before binary operators

timj:
Change W504 to W503.

cc/ @r-owen

### DIFF
--- a/python/style.rst
+++ b/python/style.rst
@@ -398,7 +398,7 @@ Error codes: E226 and E228.
 Wrap lines before binary operators
 ----------------------------------
 
-`PEP 8 does not prescribe <https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator>`_ whether lines should be broken before or after binary operators.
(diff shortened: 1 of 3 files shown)


python/style.rst
```diff
@@ -395,13 +395,13 @@ Error codes: E226 and E228.
 
 .. _style-guide-py-binary-operator-line-break:
 
-Wrap lines after binary operators
----------------------------------
+Wrap lines before binary operators
+----------------------------------
 
 `PEP 8 does not prescribe <https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator>`_ whether lines should be broken before or after binary operators.
```
ktlim:
```suggestion
`PEP 8 suggests <https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator>`_ that lines should be broken before binary operators but allows after.
```

timj:
I don't think the dev guide should talk about what happens when people hard code flake8 into their editors and the standard changes. The dev guide should describe the current status. Maybe a community post would be better for that?

ktlim:
Community seems reasonable.

r-owen:
Thank you for making this change. I wrote a community post for it:
https://community.lsst.org/t/the-rules-for-breaking-long-python-lines-is-changing/4005

PR #369, opened 2020-01-24
Replace variable with a more appropriate name

pllim:
Hello. I came across this document and thought maybe this variable should be renamed.


jonathansick:
Thank you so much for bringing this to our attention @pllim! We agree it's unacceptable we're sorry it made it into our documentation.

PR #391, opened 2020-04-21
Update git SCL's available

billglick:
add rh-git218
remove git19 (no longer maintained)
remove rh-git29 (no longer maintained)

We are only removing these from documentation for now (not necessarily removing them from dev servers).


jdswinbank:
Thanks @billglick! I fixed a minor build error here (just a typo) and merged it on #392.

PR #395, opened 2020-04-30
DM-23343: Add procedure for deprecating packages

jdswinbank:


stack/deprecating-interfaces.rst
```diff
@@ -110,3 +111,31 @@ To deprecate a `~lsst.pex.config.Field` in a `~lsst.pex.config.Config`, set the
...
 The deprecated string must also specify the version after which the config may be removed, as discussed in :ref:`code_removal`.
+
+Package Deprecation
+===================
+
+To deprecate an entire package, first have its top-level :file:`__init__.py` (e.g. :file:`python/lsst/example/package/__init__.py`) issue an appropriate `FutureWarning` when it is imported::
+
+    import warnings
+
+    warnings.warn('lsst.example.package is deprecated, and will be removed from the Rubin Observatory'
```
ktlim, line 122: No comma here since there's no second subject.  Real warnings might give a reason and a replacement, but it's OK not to have them for the example.

ktlim: If a package does not already have a top-level `__init__.py`, I guess one would have to be created.  This shouldn't happen for stack packages that have to import from `.version`, but it might happen for others.

PR #409, opened 2020-06-23
Update lsst-login.rst

billglick:
Access to login servers is now available publicly.


jdswinbank:
Hey @billglick — before this change, there were two notices on `lsst-login.rst` that these servers are under development and access is restricted. You removed one of them, but [the other is still there](https://developer.lsst.io/services/lsst-login.html#connecting-and-authenticating). Is that intentional?

PR #414, opened 2020-07-31
Update lsst-login.rst

billglick:
Move development documentation into lsst-login page, remove references to documentation under lsst-dev page.

### DIFF
--- a/services/lsst-login.rst
+++ b/services/lsst-login.rst
@@ -14,9 +14,12 @@ This page is designed to assist developers in use of the ``lsst-login`` servers:
 
 #. :ref:`lsst-login-overview`
 #. :ref:`lsst-login-connect`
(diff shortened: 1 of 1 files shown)


jdswinbank:
(Also, would you mind using a `u/billglick/...`-style user branch, please?)

billglick:
Yes, this is preparing for moving the documentation from the lsst-dev section over to lsst-login. Instead of linking to the lsst-dev documentation, we want the lsst-login documentation to be the default place for this, and will update the lsst-dev documentation to reference it on the lsst-login page. 
Should I update the lsst-dev page along with this change?

jdswinbank:
I think it would be great to update the `lsst-dev` documentation in lock-step with this — so yes please!

PR #421, opened 2020-08-04
U/billglick/20200810updates

billglick:


billglick:
Below is a summary of changes I made (not including Batch changes that @jakerundall is also including):
- update lsst-dev & lsst-dev-db to note retirement on Aug 31 2020
- reorder TOC to put less emphasis on lsst-dev & lsst-dev-db
- remove all Nebula documentation and references
- swap out references to lsst-dev to lsst-login in other pages
- update list of overall LDF resources
- remove references to lsst-login having batch submit ability

services/lsst-condor.rst
```diff
@@ -1,7 +1,15 @@
...
 ###################################################################
 Using the HTCondor DAC Cluster
 ###################################################################
 
+This content has moved: :doc:`/services/batch`.
+
+The material presented below is for historical reference and will be removed in the future.
+
+------------
+
```
billglick, line 12: @jakerundall Can we go ahead and remove all the 'content' below from this page now?

billglick:
This is probably a question for John...

The notes in the lsst-login page about git-lfs do not work for me.

> After you have initialized a shared stack, you can enable Git LFS using EUPS:
>     setup git_lfs

Should they be removed? Or the whole `git_lfs` section?

billglick:
We are planning to release these documentation updates by Monday morning (8/10) to reflect recent changes to the login and batch nodes.

jdswinbank:
> The notes in the lsst-login page about git-lfs do not work for me.
> 
> > After you have initialized a shared stack, you can enable Git LFS using EUPS:
> > setup git_lfs
> 
> Should they be removed? Or the whole `git_lfs` section removed?

These instructions only apply to the *old* shared stacks — those marked with “toolchain: devtoolset-8” (or -6) in the table above. For newer shared stacks (“toolchain: Internal (Conda)”), they aren't relevant.

I don't think we should remove this section altogether, but I would be happy to fix the wording to make that distinction more obvious. If you'd like to do so on this PR that's great; otherwise, let me know and I'll do so myself.

billglick:
> I don't think we should remove this section altogether, but I would be happy to fix the wording to make that distinction more obvious. If you'd like to do so on this PR that's great; otherwise, let me know and I'll do so myself.

@jdswinbank - I just incorporated your comments about git_lfs. You had already said most of that in the tip, but your added comments should make it even more clear.

services/batch.rst
```diff
@@ -0,0 +1,582 @@
+############################
+Using the LSST Batch Systems
+############################
+
+NCSA hosts a few `HTCondor <https://research.cs.wisc.edu/htcondor/>`_ clusters and a `Slurm <https://slurm.schedmd.com/overview.html>`_ cluster for LSST. Of particular interest to LSST DM developers will be:
+
+- ``HTCondor DAC Cluster`` - general user HTCondor pool
```
jdswinbank, line 7: I don't think there's any need to set these names in monospace. If you want to draw attention to them, probably better to use **bold**, or even a [definition list](https://developer.lsst.io/restructuredtext/style.html#definition-lists).

(Similar thing applies to repeated mention of them below).

services/batch.rst
```diff
@@ -0,0 +1,582 @@
...
+
+LSST Batch Systems: Overview
+============================
+
+THe ``HTCondor DAC Cluster`` has the following submit (schedd) nodes from which users can submit jobs and run workflows:
+
+- ``lsst-condordac-sub01.ncsa.illinois.edu``
+- ``lsst-condordac-sub02.ncsa.illinois.edu``
+
+The ``HTCondor DAC Cluster`` also has a number of dedicated compute (startd) nodes configured with partitionable slots. Commands such as those described below in :ref:`batch-condor` can be used to view the resources available in the cluster.
```
jdswinbank, line 31: ``startd`` probably does want double-backticks!

services/batch.rst
```diff
@@ -0,0 +1,582 @@
...
+
+.. _batch-stack:
+
+LSST Batch Systems: Shared Software Stack in GPFS
+=================================================
+A shared software stack on the GPFS file systems, suitable for computation on the
+``Verification Cluster``, has been provided and is maintained by Science Pipelines and
+is available under :file:`/software/lsstsw`.  This stack may be initialized via:  ::
+
+     % .  /software/lsstsw/stack/loadLSST.bash
```
jdswinbank, line 89: I suggest using [this syntax](https://developer.lsst.io/restructuredtext/style.html#command-line-prompts) to make your example prompts match those elsewhere in the documentation.

If you _don't_ use that syntax, please change the line above to be:

> ... initialized via::

Which should have the same effect but be a bit less verbose than what you have now!

services/lsst-condor.rst
```diff
@@ -1,7 +1,13 @@
+:orphan:
```
jdswinbank, line 1: I think this file should just be removed if it is orphaned; I don't see any value to keeping it around.

team/onboarding.rst
```diff
@@ -43,16 +43,10 @@ Data Facility resources
 
 You'll also need an account for the LSST Data Facility services, hosted at NCSA, which include:
 
-- Development Servers (`lsst-dev </services/lsst-dev.html>`_)
-- MySQL Database (`lsst-dev-db </services/lsst-db.html>`_)
+- Development Servers (`lsst-dev </services/lsst-login.html>`_)
```
jdswinbank, line 46: At least for the time being, I think it would be appropriate to mention the ``lsst-dev`` systems here: I understand from @mbutlerncsa that they are not being retired quite yet, and not all of the services that they currently provide are yet available elsewhere.

billglick: This talks about accounts. These are the same accounts. Listing both implies different accounts to me.
We point new users to lsst-login and provide information about lsst-dev and other resources

services/lsst-condor.rst
billglick: We have current HTCondor users and testers that use this file for reference. I don't think we just want to yank it out from under them (at least not yet).

billglick:
fyi... Nebula was removed as an LSST resource back in February 2020. I'm not sure how this was announced -- I'm just the messenger on that one.

billglick:
I swapped many generic references to LSST to be Rubin Observatory, etc. The main things I skipped were other things, e.g.:
- LSST Stack
- LSST Software
- and obviously hostnames

team/onboarding.rst
billglick: Actually the new documentation does say lsst-dev, but points to the documentation that was moved under lsst-login. That might be confusing, but is probably less confusing for newish developers that may have been told that they need an 'lsst-dev' account. I think we leave it as revised here.

PR #427, opened 2020-08-25
Add system python 2.7 & 3.6 documenation

billglick:
Triggered by https://jira.lsstcorp.org/browse/IHS-4152


services/software.rst
```diff
@@ -77,6 +77,7 @@ In addition to developer tools provided by the :ref:`software-stack`, the follow
 - **Compilers**: ``c``, ``c++``, and ``fortran`` from gcc 4.8.5
 - **Debuggers**: ``glibc-debuginfo`` (i.e. ``gdb``)
 - **Build Tools**: ``autoconf``, ``automake``, ``bison``, ``blas``, ``byacc``, ``cmake``, ``flex``, ``fontconfig``, ``make``, ``valgrind``, ``yum-utils``, etc.
+- **Python**: ``python2`` / ``pip2`` (v2.7) and ``python3`` / ``pip3`` (v3.6)
```
timj, line 80: I'd be inclined to not mention python2 anywhere if we could avoid it.

billglick: The reason that I mention it is that the system `python` command references `python27`. But, I'm fine just striking the python2/pip2 part of this:
```suggestion
- **Python**: ``python3`` / ``pip3`` (v3.6)
```
Is that better?

PR #428, opened 2020-08-26
DM-24398: Add Click documentation to the developer guide

n8pease:


python/cli.rst
```diff
@@ -0,0 +1,68 @@
...
+1. ``SConscript`` with contents:
+
+.. code-block:: py
+
+    # -*- python -*-
+    from lsst.sconsUtils import scripts
+    scripts.BasicSConscript.shebang()
+
+2. A file that has the name of the CLI command the user will call. This file should contain as little
+   implementation as possible, it should import an implementation file and call a function in it. For example:
```
ktlim, line 43:
```suggestion
2. A file that has the name of the CLI command the user will call.
This file should contain as little implementation as possible.
It should import an implementation file and call a function in it.
For example:
```
Maybe also explain why there should be no implementation.

python/cli.rst
```diff
@@ -0,0 +1,68 @@
+##############################
+Python Command Line Interfaces
+##############################
+
+Command line interfaces (CLI) in Python may be implemented using the ``argparse`` module or using `Click`_.
```
jdswinbank, line 5: Why not link ``argparse`` to https://docs.python.org/3/library/argparse.html?

jdswinbank: Can you add a comment here as to why one choose Click over argparse, or vice versa? It'd be great if the developer reading this page could receive clear guidance about what they ought to do, rather than just a list of the possibilities.

python/cli.rst
```diff
@@ -0,0 +1,68 @@
...
+Click is a Python package for creating command line interfaces in a composable way. It supports:
+
+* arbitrary nesting of commands
+* automatic help page generation
+* lazy loading of subcommands at runtime
+
+`Click`_ has good `documentation`_ including a `quickstart`_ guide.
+
+.. _documentation: https://click.palletsprojects.com/en/7.x/#documentation
+.. _quickstart: https://click.palletsprojects.com/en/7.x/quickstart/
```
jdswinbank, line 23: A little unfortunate to hard-code a link to v7 documentation, but maybe there's no good alternative (since `master` docs have a big warning label across the top). Hohum! :-)

python/cli.rst
```diff
@@ -0,0 +1,68 @@
...
+
+The ``butler`` CLI command is implemented in ``daf_butler``. It has a reusalbe framework for dynamically
```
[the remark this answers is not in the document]
jdswinbank: As K-T implies, it's not really why this is here; the implementation of daf_butler doesn't seem to be particularly germane to the topic at hand. (Our C++ docs don't bother mentioning that Qserv uses C++, for example!)

If this intended just to be a “see daf_butler for an example of where Click has been used”, please make that clear (per K-T).

If it's intended to be more than that — perhaps you are advocating that other packages should actually import and use the daf_butler framework directly — please be explicit. And link to the daf_butler docs directly (they're over [here](https://pipelines.lsst.io/modules/lsst.daf.butler/index.html)). And consider whether we really want a reusable CLI framework in the daf_butler package — it seems like maybe we should file a ticket for making it an independent library.

python/cli.rst
```diff
@@ -0,0 +1,68 @@
...
+    A file ``bin.src/calibrate.py`` is implemented as:
+
+    .. code-block:: py
+
+        from lsst.pipe.tasks.calibrate import CalibrateTask
+
+        CalibrateTask.parseAndRun()
+
+
+
```
jdswinbank, line 68: Please remove the redundant white space.

python/cli.rst
```diff
@@ -0,0 +1,68 @@
...
+The ``butler`` CLI command is implemented in ``daf_butler``. It has a reusalbe framework for dynamically
+loading subcommands from plugin packages. It is documented in the ``daf_butler`` package documentation.
+
+Add a Callable CLI Command To Your Package
+==========================================
+
+To create a callable command at the top level of your package create a folder called ``bin.src``, it should
+contain two files:
+
+1. ``SConscript`` with contents:
```
jdswinbank, line 34: The indentation in this list is a little off, which makes the rendering look wrong — in the attached screenshot, compare the version on the left with my re-indented version on the right. 
Screenshot 2020-08-27 at 10 32 03

The diff is:
```
diff --git a/python/cli.rst b/python/cli.rst
index 5f9939f..4dcfe0b 100644
--- a/python/cli.rst
+++ b/python/cli.rst
@@ -33,36 +33,33 @@ contain two files:

 1. ``SConscript`` with contents:

-.. code-block:: py
+   .. code-block:: py

-    # -*- python -*-
-    from lsst.sconsUtils import scripts
-    scripts.BasicSConscript.shebang()
+       # -*- python -*-
+       from lsst.sconsUtils import scripts
+       scripts.BasicSConscript.shebang()

 2. A file that has the name of the CLI command the user will call. This file should contain as little
    implementation as possible, it should import an implementation file and call a function in it. For example:

-    a. **Click Scripts**
+   a. **Click Scripts**

-    A file ``bin.src/butler`` is implemented as:
+      A file ``bin.src/butler`` is implemented as:

-    .. code-block:: py
+      .. code-block:: py

-        import sys
-        from lsst.daf.butler.cli.butler import main
+          import sys
+          from lsst.daf.butler.cli.butler import main

-        if __name__ == '__main__':
-            sys.exit(main())
+          if __name__ == '__main__':
+              sys.exit(main())

-    b. **Pipe Task Scripts**
+   b. **Pipe Task Scripts**

-    A file ``bin.src/calibrate.py`` is implemented as:
-
-    .. code-block:: py
-
-        from lsst.pipe.tasks.calibrate import CalibrateTask
-
-        CalibrateTask.parseAndRun()
+      A file ``bin.src/calibrate.py`` is implemented as:

+      .. code-block:: py

+          from lsst.pipe.tasks.calibrate import CalibrateTask

+          CalibrateTask.parseAndRun()
```

python/cli.rst
```diff
@@ -0,0 +1,68 @@
...
+
+    .. code-block:: py
+
+        import sys
+        from lsst.daf.butler.cli.butler import main
+
+        if __name__ == '__main__':
+            sys.exit(main())
+
+    b. **Pipe Task Scripts**
```
jdswinbank, line 57: Why “Pipe Task”? Is the implication that Click is to be used for all scripts *except* those which are calling `parseAndRun()` on a task? If so, could you clarify that above?

In general, it's not really clear whether this section is describing requirements (“it *must* be done this way”) or suggestions (“you could do it like this if you like”).

PR #429, opened 2020-08-27
DM-21147: Permit snake_case variables and functions more broadly.

TallJimbo:
This implements RFC-623.

### DIFF
--- a/cpp/style.rst
+++ b/cpp/style.rst
@@ -195,16 +195,19 @@ The capitalization rule for class names should be all words in the name capitali
 
 .. _style-guide-cpp-3-2:
 
-3-2. Variable names MUST be in mixed case starting with lower case.
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+3-2. Variable names MUST be camelCase or snake_case.
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: cpp
 
    int lineWidth;
+   // OR, in primarily new code:
+   int line_width;
 
-Common practice in the C++ development community.
-Makes variables easy to distinguish from types, and effectively resolves potential naming collision as in the declaration Line line.
-Keep variable names balanced between short and longer, more meaningful.
+While the broader C++ community has not strongly converged on either of these styles, the Python community has converged towards snake_case while much of our code is historically camelCase, and consistency between our C++ and Python code is our priority in naming conventions.
+See :ref:`style-guide-py-naming` for guidelines for when to use each style.
+
+In general, keep variable names balanced between short and longer, more meaningful.
 Use 8 to 20 characters as a guideline (excluding integer loop counters which may be as little as 1 character).
 
 .. _style-guide-cpp-3-3:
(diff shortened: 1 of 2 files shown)


cpp/style.rst
```diff
@@ -195,17 +195,17 @@ The capitalization rule for class names should be all words in the name capitali
...
    int lineWidth;
+   // OR, in primarily new code:
+   int line_width;
 
-Common practice in the C++ development community.
-Makes variables easy to distinguish from types, and effectively resolves potential naming collision as in the declaration Line line.
-Keep variable names balanced between short and longer, more meaningful.
-Use 8 to 20 characters as a guideline (excluding integer loop counters which may be as little as 1 character).
+While the broader C++ community has not strongly converged on either of these styles, the Python community has converged towards snake_case while much of our code is historically camelCase, and consistency between our C++ and Python code is our priority in naming conventions.
+See :ref:`style-guide-py-naming` for guideliness for when to use each style.
```
ktlim, line 208:
```suggestion
See :ref:`style-guide-py-naming` for guidelines for when to use each style.
```

cpp/style.rst
```diff
@@ -195,17 +195,17 @@ The capitalization rule for class names should be all words in the name capitali
 
 .. _style-guide-cpp-3-2:
 
-3-2. Variable names MUST be in mixed case starting with lower case.
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+3-2. Variable names MUST be camelCase or snake_case.
```
ktlim, line 198: I guess we don't need to explicitly say "start with lower case"?

cpp/style.rst
```diff
@@ -195,17 +195,17 @@ The capitalization rule for class names should be all words in the name capitali
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: cpp
 
    int lineWidth;
+   // OR, in primarily new code:
+   int line_width;
 
-Common practice in the C++ development community.
-Makes variables easy to distinguish from types, and effectively resolves potential naming collision as in the declaration Line line.
```
ktlim, line 206: Is it worth keeping this explanation?

cpp/style.rst
```diff
@@ -195,17 +195,17 @@ The capitalization rule for class names should be all words in the name capitali
...
-Keep variable names balanced between short and longer, more meaningful.
-Use 8 to 20 characters as a guideline (excluding integer loop counters which may be as little as 1 character).
```
ktlim, line 208: Is it worth keeping this suggestion?

python/style.rst
```diff
@@ -511,15 +511,28 @@ For historical reasons, Science Pipelines code (nominally, all packages included
...
+  "snake_case" local variables names are similarly preferred in new functions, even functions with a camelCase public API.
+  The strict definition of "primarily new code" is case by case and left to developer judgement.
+
+- camelCase is still permitted in new code, and is preferred when adding public methods, arguments, and attributes to existing classes with an established camelCase API, or making modifications to existing functions with camelCase local variables.
+
+- Under no circumstances should the Python side of a pybind11-wrapped C++ function use a different naming convention than the C++ function.  C++ naming conventions and the guidelines for applying them in new/old code are no different from those of Python, but it is never permitted to just change things at the pybind11 level; either both C++ and Python or neither should be changed.
+
+Changing existing names from camelCase to snake_case is generally discouraged as unnecessary churn, and should only be done in highly localized code (e.g. individual function bodies) that already being extensively modified.
```
ktlim, line 534:
```suggestion
Changing existing names from camelCase to snake_case is generally discouraged as unnecessary churn, and should only be done in highly localized code (e.g. individual function bodies) that is already being extensively modified.
```

python/style.rst
```diff
@@ -511,15 +511,28 @@ For historical reasons, Science Pipelines code (nominally, all packages included
...
+Name changes to public interfaces are of course API changes, and must go through the usual RFC and deprecation procedure.
```
ktlim, line 535:
```suggestion
Name changes to public interfaces are of course API changes and must go through the usual RFC and deprecation procedure.
```

ktlim (approved):

cpp/style.rst
TallJimbo: My sense of this - and the other block of reasoning I removed below - was that in this aspect our C++ style guide is now just a reflection of our Python style guide, because Python is now really our primary language and there's no good reason for the languages to differ on this aspect of style.  I'm sure the same general-best-practice arguments are ultimately behind the Python style choices, too, but as we delegate those "why" questions in Python to PEP8, it felt weird to continue enumerating them here.

cpp/style.rst
TallJimbo: I don't think it's likely that someone would incorrectly guess that starting with caps is okay, given the examples here and throughout the codebase, and I think the reference to the Python style guide and PEP8 provides clarification for anyone looking hard for a clear ruling.

cpp/style.rst
TallJimbo: I've restored the last two bullets, as those are indeed suggestions; I'd missed the distinction in my first read-through of these comments.

PR #430, opened 2020-08-27
DM-25710: Clarify that executable scripts do not need .py suffixes.

TallJimbo:
...but they should be minimal, especially if they don't have a .py
suffix.

This implements RFC-698.

### DIFF
--- a/python/style.rst
+++ b/python/style.rst
@@ -619,7 +619,9 @@ Test files must have the form ``test_{description}.py`` for compatibility with P
 The name of a test case should be descriptive without the need for a trailing numeral to distinguish one test case from another.
 
 This rule does not apply to executable script files, for which both no extension and a '.py' extension are acceptable.
-Script files should always be minimal (ideally a single non-import statement), and delegate any actual logic to importable code; this means the impact of having no extension on tools that rely on the file extension should be negligible.
+Script files should always be minimal (ideally a single non-import statement), and delegate any actual logic to importable code.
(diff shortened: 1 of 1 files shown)


python/style.rst
```diff
@@ -605,6 +605,10 @@ A module containing a single class should be a ``camelCase``-with-leading-lowerc
 Test files must have the form ``test_{description}.py`` for compatibility with Pytest.
 The name of a test case should be descriptive without the need for a trailing numeral to distinguish one test case from another.
 
+This rule does not apply to executable script files, for which both no extension and a '.py' extension are acceptable.
+Script files should always be minimal (ideally a single non-import statement), and delegate any actual logic to importable code; this means the impact of having no extension on tools that rely on the file extension should be negligible.
```
ktlim, line 609: Might be good to say _why_ we prefer minimal script files.

ktlim (approved):

python/style.rst
TallJimbo: I've updated this to say that minimal script files are better because they maximize our ability to reuse code from Python, not just the command line.  That's what I think of as the motivation, but it occurs to me that others may have other reasons that they consider equally or more important.

FWIW, it also occurs to me that our policy of putting most unit tests in decidedly non-minimal scripts may not really be a good one, though that's certainly not something I want to challenge on this ticket.

timj: Minimal scripts makes unit testing much easier (since you can ignore trying to use subprocess) and integrates the command easily into the sphinx documentation (which bin.src makes hard).

PR #439, opened 2020-09-23
Update kinit usage in lsst-login page.

ktlim:


RobertLuptonTheGood (commented):
Looks good, thank you

arunkannawadi:
Is it worth including that one could see when the ticket expires with `klist`?

services/lsst-login.rst
```diff
@@ -76,16 +76,19 @@ The Kerberos domain for the ``lsst-login`` servers is ``NCSA.EDU``, so something
...
 
+  # On a Mac, your Kerberos password may be saved in your keychain.
+  # Use /usr/bin/kinit (instead of any conda kinit you may have in your path) and specify "--keychain" to save.
+  
 .. tip::
 
    **Kerberos Tickets Expire**
 
-   - Your Kerberos ticket on your local machine will expire (generally 25 hours after inititally granted) and need to be renewed, which you can do with ``kinit -R``.
+   - Your Kerberos ticket on your local machine will expire (by default 10 hours but up to 25 hours after initially granted with the ``-l`` option shown above) and need to be renewed, which you can do with ``kinit -R``.
```
billglick, line 91: The default depends on your local kerberos config file. So, there isn't really a default.
```suggestion
   - Your Kerberos ticket on your local machine will expire (e.g. 25 hours after initially granted with the ``-l`` option shown above) and need to be renewed, which you can do with ``kinit -R``.
```

billglick: I guess the default _is_ actually 10 hours, but this can be configured for up to 25 hours in the `kerb5.conf` file. So, the original is fine.

PR #440, opened 2020-09-25
September 2020 staff updates

jdswinbank:


jdswinbank:
I agonised briefly about assigning copyright to ASTRON, or to me directly — I don't know if it's more likely that future work would be on my own time or on my employer's, and I don't know if they'd assert a copyright claim to something I did on my own. For now, I think this is fine, and we can revisit if it ever actually becomes an issue.

timj:
> I agonised briefly about assigning copyright to ASTRON, or to me directly — I don't know if it's more likely that future work would be on my own time or on my employer's, and I don't know if they'd assert a copyright claim to something I did on my own. For now, I think this is fine, and we can revisit if it ever actually becomes an issue.

Surely if you are doing work it's only going to be owned by ASTRON if they are asking you to do that work.

PR #446, opened 2020-10-05
Replace John with Yusra

arunkannawadi:
As Yusra AlSayyad has taken over the role of John Swinbank, the DRP page is updated to show this change.


jdswinbank (approved):
Looks good! Thanks.

PR #448, opened 2020-10-08
Focus

womullan:


ktlim (approved):
I think for better URLs and SEO it would be good to name the page `focus-friday` instead of just `focus`.  Otherwise looks fine.

womullan:
name changed

PR #450, opened 2020-10-27
Fix dead links.

ktlim:
As reported by @cwwalter.


timj (approved):
I suppose there's no reason to copy the trac page over?

PR #454, opened 2020-11-18
Remove double https

arunkannawadi:


arunkannawadi:
I just edited it directly on GitHub and not sure why the build fails. It's probably easier if any one from the docs team fixes it directly instead.

ktlim:
The failure is due to https://jira.lsstcorp.org/browse/DM-27536.  Not your fault.  Thanks for the fix.

ktlim (approved):
This whole page may disappear if I can manage to automate the process, but thanks for the fix.

PR #455, opened 2020-11-24
Slight relaxation on slack use

womullan:


team/focus-friday.rst
```diff
@@ -12,7 +12,7 @@ By reducing traffic we will reduce the need to participate in threads on specifi
 Hence there should be: 
 
 - No planned meetings, generally. There may of course be exceptions, for example if your focused work includes a co-work with others who agree to meet, this is OK. Such agreements should be between peers and not use any hierarchy. You may elect to attend meetings/workshops outside DM — or occasionally we may be required to attend such meetings.
-- No Slack messages on general channels. You may slack an individual, or a group of individuals if you have agreed such interactions ahead of time. Conversations should not be carried out on general channels. There should be no expectation of response to Slack messages.
+- Slack messages on general channels are discouraged and one should not expect a response. Conversations are discouraged on general channels and specific users should not be @ed. You may slack an individual, or a group of individuals if you have agreed such interactions ahead of time. There should be no expectation of response to Slack messages.
```
RobertLuptonTheGood, line 15: Can you define a "general channel"?  Is it OK to ask for help with a service that isn't working on a Friday (e.g. RSP problems with logging in, which is often a bad pod)?

ktlim: General == non-private, I think.

See the next section for emergencies/bugs/failures.

ktlim: "There should be no expectation of response to Slack messages." is now redundant with "one should not expect a response".

SimonKrughoff: As Frossie said, we do monitor #dm-rsp and similar service specific channels, though please think about whether the support you need is actually blocking you or whether it can wait until Monday

RobertLuptonTheGood: Yes, but that was at DMLT not in this document.  I think we should be clear about support.  "I can't get my work done today" seems reasonable to me --- we're trying to maximise the productivity on Fridays for DM, and I'm afraid that this'll mean some people having to provide support rather than get their own work done.

RobertLuptonTheGood: If "general" means "public" we should use that word.

SimonKrughoff: I want to emphasize that it's up to you what getting your work done means.  You may find yourself in a position of not being able to get a specific thing done on Friday to ease support for someone else.  In that case you may need to pivot to something else.

I do understand some things are time critical and simply can't wait.

frossie: It is always okay to report a service not working on a support channel at any time of day or night.

womullan: adding under emergencies This may include a post to public slack support channel.

ktlim (approved):
We don't have a great way of identifying "support" or "general discussion" channels, but hopefully people can figure it out for themselves.

PR #456, opened 2020-11-24
Add tip about using passcodes for Duo auth

SimonKrughoff:


ktlim (approved):

services/lsst-login.rst
```diff
@@ -62,6 +62,15 @@ Information on setting up NCSA Duo is available at the following URL:
    Only password (keyboard-interactive) and Kerberos (gssapi) authentication are allowed via SSH on the lsst-login nodes.
 
 
+.. tip::
+
+   **Using passcodes**
+
+   When logging in, you will be prompted for the device on which you wish the Duo push notice to arrive.
+   Instead of selecting a device, you may, instead, enter the six digit passcode from your Duo app.
+   This can be useful if, for whatever reason, your device cannot recieve push notices.
```
arunkannawadi, line 71: Worth mentioning that the passcode is valid only for 30 seconds?

SimonKrughoff: OK, will do

PR #462, opened 2020-12-17
Make resource limits on lsst-login more obvious

parejkoj:


services/lsst-login.rst
```diff
@@ -22,12 +22,9 @@ This page is designed to assist developers in the use of the ``lsst-login`` serv
...
 ========
 
-The ``lsst-login`` servers are primarily intended as bastions used to access other resources at NCSA. Additional capabilities include:
-
-- light development with short-running processes that require modest resources (e.g., building docs, short compilations against LSST software stack)
-- view files (e.g., FITS files)
-
-Users are encouraged to submit batch jobs to perform work that requires more significant resources. Please see :doc:`/services/batch` for more information.
+The ``lsst-login`` servers are primarily intended as bastions used to access other resources at NCSA.
+This node should only be used as a proxy for lsst-devl (see :doc:`/services/lsst-devl`) or submit batch jobs to perform work that requires more significant resources (see :doc:`/services/batch`).
```
ktlim, line 26:
```suggestion
These servers should only be used as a jump host for the ``lsst-devl`` machines (see :doc:`/services/lsst-devl`) or to submit batch jobs to perform work that requires more significant resources (see :doc:`/services/batch`).
It may also be easier to use these machines than the ``lsst-devl`` machines with remote editors and IDEs.
```

services/lsst-login.rst
```diff
@@ -22,12 +22,9 @@ This page is designed to assist developers in the use of the ``lsst-login`` serv
...
+This node has resource limits that will cause larger tasks (such as compiling afw) to fail due to memory and other limits.
```
ktlim, line 27:
```suggestion
These servers have resource limits, such as a small amount of memory per core, that will cause larger tasks (such as compiling ``afw``) to fail.
```

ktlim (approved):
There are multiple login machines, so "This node" is inappropriate.  A few other wording suggestions, plus one other possible use.

PR #463, opened 2020-12-22
Text will no longer be in LDM-294 (RFC-747)  this  is now the source

womullan:


womullan:
I think this could go ahead

ktlim (approved):
Looks good.

PR #469, opened 2021-02-26
Update ruler from 80 to 79

arunkannawadi:
The dev guide requires restricting docstring and comments to 79 characters, and hence a ruler at 79 instead 80 makes the most sense. The equivalent config for Sublime Text also says 79.


leeskelvin (approved):

jonathansick:
@arunkannawadi For future reference, we don't merge `master` into our ticket branches in order to update ticket branches.

Instead, always rebase your branch against the latest `master` branch. See https://developer.lsst.io/work/flow.html#merging

arunkannawadi:
Yes, apologies John. I was editing it directly on Github and pressed the green button to merge without reading the message. It was too late before I realized that I had merge `master` into my branch. Let me know if I should edit the Git history.

PR #471, opened 2021-03-17
Remove kerberos requirement for HTCondor

billglick:


billglick:
@MichelleGower reviewed this and communicated the following to me:

> It looks fine.  I tried searching for Kerberos in the developer guide and didn't see any other places related to htcondor.

PR #475, opened 2021-04-02
DM-29027: Update to C++17.

ktlim:


cpp/boost.rst
```diff
@@ -12,11 +12,14 @@ Using Boost
 
 A Boost library may be used only if:
 
-1. the desired effect cannot be accomplished with a C++14 standard language feature, and
+1. the desired effect cannot be accomplished with a C++17 standard language feature, and
 2. a C++ standard library equivalent is either unavailable or unusable with our minimum required compiler version (i.e. ``gcc`` version 6.3.1).
```
erykoff, line 16: I don't know where we specify the minimum gcc or clang version (and this is less relevant since we're on conda compilers) but we need gcc >= 7 for "full" c++17 features including those listed in the updates below (`std::any`, `std::variant`, etc.): https://gcc.gnu.org/projects/cxx-status.html and https://gcc.gnu.org/gcc-7/changes.html#cxx

erykoff (approved):
Looks good with the caveat about gcc versions.

cpp/boost.rst
ktlim: It is specified indirectly in the Style Guide as coming from rubin-env.  This is currently at gcc 9.3.0.  I've put a pointer to the Style Guide in this line.

PR #480, opened 2021-04-21
DM-29838: Note that LSST Corp no longer owns any copyrights

timj:


legal/copyright-overview.rst
```diff
@@ -54,11 +54,12 @@ LSST Corporation and preconstruction era
 ----------------------------------------
 
 Prior to the start of LSST construction (before August 2014) all copyright was granted to the LSST Corporation.
+LSST Corporation `have transferred <http://ls.st/agreement-221>`_ those copyrights to AURA and any remaining references to the LSST Corporation owning copyright should be corrected.
 
 .. note::
 
    Since the beginning of construction (August 2014) some copyrights have been erroneously granted to the LSST Corporation because of a lack of policy communication.
-   These post-2014 copyright assignments to LSST Corporation are errors and should be fixed.
+   Any remaining statement of code being owned by the LSST Corporation should be changed to AURA.
```
gpdf, line 62: I think that means that the example at line 82 below should be changed?  Perhaps, since the point is a two-institution example, it could show, e.g., AURA and Caltech?

gpdf (changes requested):
I think the example needs to be fixed or the guidance may still be read as ambiguous.

legal/copyright-overview.rst
timj: I fixed it. Thanks.

gpdf (approved):
This looks good to me now.  Thank you!

womullan (approved):
looks good

PR #481, opened 2021-04-26
Add a note about direct messages

arunkannawadi:


communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
 -  ``:thank_you:`` (🙏) or ``:thankyou:``
 
+.. _slack-direct-messages:
+
+Direct messages
+---------------
+
+Please keep direct messaging to a user or a set of users to minimum.
+Direct messages should be avoided for any technical discussions or asking for support, even if you are sure that you have involved all those who are interested or who can contribute.
+Instead, such discussions should be in a thread in an appropriate :ref:`discussion channel<slack-discussion-channels>`.
```
ktlim, line 93: Support requests should be in a support channel.

communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
+Direct messages shoul be reserved for conversations on topics such as code review requests, logistics, or on other topics not directly related to work.
```
ktlim, line 94: "shoul" → "should".  The phrasing here is not great ("topics" twice, no "on" with logistics, "reserved" but a non-exhaustive list).  It might be better to be more explicit with reasons and considerations (pulling them out of the Community post) rather than trying to be prescriptive.  And the fact is that there are many, many direct messages anyway, so if we are describing the existing culture, this seems inaccurate.  (I think it is OK to try to gently nudge the culture in a different direction, but it's more difficult to overturn it, and that should probably be via RFC as well.)

communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
+See this `community post`_ for why direct messages are considered harmful.
+
+.. _community post: https://community.lsst.org/t/dms-in-slack-considered-harmful/4036
+
+.. _slack-editing:
+
+Editing messages
```
ktlim, line 101: Include "deleting" in the header as well?

communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
+----------------
+
+Slack has a feature that allows a user to edit a message sent by that user.
+This feature must be used for  typographical or grammatical errors that the user wishes to correct, or if the message is accidentally sent  before it is fully typed out.
```
ktlim, line 105: "Must be used" seems too strong, but "wishes" then makes it seem much weaker.  Can you rephrase this?  Extra space.

communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
+While Slack is less formal than email, messages should be coherent and in multiple sentences if required, instead of multiple single-line messages in short bursts. This is so that it is easy to link to them elsewhere in the future.
```
ktlim, line 106: Again, I think it might be better to lead with the reasons and considerations and de-emphasize the prescriptiveness.

Generally you should give each sentence its own line.

communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
+In place of amending a message, it is preferable to use strikethrough to retract an idea or statement that If you post something you recognize as wrong.
```
ktlim, line 107:
```suggestion
In place of amending a message, it is preferable to use strikethrough ``~text~`` to retract an idea or statement if you post something you recognize is wrong.
```
Maybe say why?

communications/slack-culture.rst
```diff
@@ -83,6 +83,31 @@ You can save space, time, and distractions to others by using commonly understoo
...
+A corrected statement can be added next to it.
+A message should be deleted only if posted in a wrong channel, or if it is not appropriate in a social forum.
```
ktlim, line 109: Why?  If it may be confusing to others who come after, isn't removing it altogether better than striking through it?

communications/slack-culture.rst
```diff
@@ -191,7 +216,7 @@ If you need to have a group created because you have a routine need to @-mention
 @-mentioning a particular user notifies that user and adds the message to their "Mentions & reactions" list.
 Be aware of the person's timezone before @-mentioning them.
 If you need to @-mention them outside normal hours, you may want to schedule the message to be posted later (using the ``/send`` or more sophisticated ``/schedule`` shortcuts from `Timy`_).
-Sometimes @-mentioning a user can be helpful to disambiguate replies in a multi-person conversation, but if you find yourself prefixing every message you send with an @-mention, consider whether it is implied by context or whether the whole conversation might be better in a thread or in direct messages to that user.
+Sometimes @-mentioning a user can be helpful to disambiguate replies in a multi-person conversation, but if you find yourself prefixing every message you send with an @-mention, consider whether it is implied by context or whether the whole conversation might be better in a thread.
```
ktlim, line 219: I am OK with this, but would you consider a compromise that would de-emphasize the direct messages while leaving them as a possibility?

communications/slack-culture.rst
arunkannawadi: I find it difficult to lead with reasons always, but please feel free to modify it if you find a better way of phrasing.

communications/slack-culture.rst
arunkannawadi: Is this comment for strikethrough or deleting? I think deleting would be confusing since I might see the message in my notification but not on Slack. Strikethrough keeps a record of what was said, and that it was taken back.

communications/slack-culture.rst
arunkannawadi: I modified to say direct messages would be acceptable if the discussion is unique to a user, and give an example.

communications/slack-culture.rst
arunkannawadi: I thought the point of the dev guide is not to describe what the current culture is, but rather serve as a guide to what the culture ought to be. If we, as a team, indeed want to keep DMs to a minimum, then whenever we find ourselves in DMs that are getting technical, we can point to the dev guide and ask to move the conversation in an appropriate channel.

arunkannawadi: Are direct messages happening a lot within the data management team? The [analytics](https://lsstc.slack.com/stats) that Eric had posted on the community post includes Science Collaborations and other parts of the project. I personally don't use a lot of direct messaging, and my impression was that data management in general uses less direct messages than other users in the workspace.

PR #482, opened 2021-04-28
DM-29889: Update documentation for building Science Pipelines documentation with Documenteer in conda

jonathansick:
This PR updates the Pipelines documentation build instructions given that documenteer is now included in the Science Pipeline's conda environment:

- https://developer.lsst.io/v/DM-29889/stack/building-single-package-docs.html
- https://developer.lsst.io/v/DM-29889/stack/building-pipelines-lsst-io-locally.html

In https://github.com/lsst/templates/pull/118 we're updating the template, and that will automatically be reflected in an updated section on the doc/conf.py file in the "Layout of the doc/ directory page" https://developer.lsst.io/stack/layout-of-doc-directory.html#the-doc-conf-py-file (the code sample pulls data from the templates repo).

### DIFF
--- a/stack/building-pipelines-lsst-io-locally.rst
+++ b/stack/building-pipelines-lsst-io-locally.rst
@@ -40,12 +40,6 @@ Then set up the `pipelines_lsst_io`_ package with EUPS:
 
    setup -r pipelines_lsst_io
 
-.. warning::
-
-   If you’ve already have packages set up with the :command:`setup` command, you might need to un-setup them with the :command:`unsetup` command before running ``setup -r pipelines_lsst_io``.
-
-   `pipelines_lsst_io`_ acts as a top-level EUPS package, and its table file defines what packages are included in the `pipelines.lsst.io`_ documentation site.
-
 .. _local-pipelines-lsst-io-build-documenteer:
 
 Install Documenteer, the documentation tooling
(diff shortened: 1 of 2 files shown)


stack/building-pipelines-lsst-io-locally.rst
```diff
@@ -20,6 +20,10 @@ Before starting, you’ll need a working ``lsst_distrib`` installation.
 
 This installation needs to be a recent daily or weekly build so that any in-development packages will compile with the Stack. Working from the tip of the ``master`` branch is the norm for LSST software development.
 
+Finally, the documentation build uses Documenteer_ and related Sphinx_ documentation packages.
+Documenteer_ is already installed if you are using the Rubin Conda environment (part of the usual Science Pipelines installation).
```
ktlim, line 24: This text is a bit problematic, as the documenteer version installed with the stack is not yet operational for building pipelines.lsst.io, as I understand it.  We are still building in Jenkins with 0.5.x from `requirements.txt`, not 0.6.x from conda, although we're close to changing over.

jonathansick:
Would you prefer we hold back on the update to `stack/building-pipelines-lsst-io-locally.rst` and instead only update `/stack/building-single-package-docs.rst` (which is the update that our team is most vocal about).

ktlim:
Yes, I think that works.

stack/building-pipelines-lsst-io-locally.rst
```diff
@@ -40,12 +40,6 @@ Then set up the `pipelines_lsst_io`_ package with EUPS:
 
    setup -r pipelines_lsst_io
 
-.. warning::
```
ktlim, line 43: Does it make sense to say that if you have locally-modified clones setup that you should use `-k` to keep them setup?

PR #483, opened 2021-05-05
Change ssh config for devl to use ProxyCommand

parejkoj:
This is the config I was able to get to work on my mac laptop and linux desktop. The other method never worked for direct "ssh lsst-devl01" connections.

### DIFF
--- a/services/lsst-login.rst
+++ b/services/lsst-login.rst
@@ -161,17 +161,17 @@ A relatively complete ``~/.ssh/config`` "recipe" for streamlining your SSH conne
    # Define an alias and config for the internal nodes, which can only be reached through a login node.
    Host lsst-devl*
       User NCSAUSERNAME
-      ForwardAgent yes
+      ForwardAgent yes  # To allow Github ssh without re-authentication.
       ServerAliveInterval 120
    Host lsst-devl01
       HostName lsst-devl01.ncsa.illinois.edu
-      ProxyCommand ssh lsst-login01 -W %h:%p
+      ProxyJump lsst-login01
    Host lsst-devl02
       HostName lsst-devl02.ncsa.illinois.edu
-      ProxyCommand ssh lsst-login02 -W %h:%p
+      ProxyJump lsst-login02
    Host lsst-devl03
       HostName lsst-devl03.ncsa.illinois.edu
-      ProxyCommand ssh lsst-login03 -W %h:%p
+      ProxyJump lsst-login03
 
 With such config in ``~/.ssh/config`` on your local machine, your SSH connections can be significantly streamlined. Your experience may look like this:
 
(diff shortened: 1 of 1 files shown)


rra (commented):
This removes all the GSSAPI configuration which will break folks who are using Kerberos if they adopt this. Was that critical to making it work for you?

The change from `ProxyJump` to `ProxyCommand` implies to me that you have an ancient version of SSH that doesn't support `ProxyJump`. Is that the case? `ProxyJump` is preferred to `ProxyCommand` if it exists (and does the same thing, except better).

Enabling agent forwarding to those servers is a mild increase in security risk but probably a substantial benefit in usability for things like GitHub, so seems reasonable to me.

ktlim:
I think this is basically a "I use Kerberos" versus "I refuse to use Kerberos" debate, so the solution is probably to have both configs rather than try to replace the one that's there.

ktlim:
(Branch names in this repo should still follow DM workflow standards, i.e. `u/parejkoj/ssh-config-ProxyCommand`.)

TallJimbo:
FWIW, old configs (more or less) and Kerberos have always worked for me.  Forwarding the SSH agent is something that I stopped doing when passphrase-based SSH stopped being possible at NCSA, because it stopped Just Working (may have been fixable and I was just lazy); I've switched to HTTPS/personal-access-token authentication for my git repos at NCSA.

parejkoj:
Thank you for the responses. I'm a non-Kerberos user. I dug into it a bit more:

* ProxyJump works fine, so I've replaced the ProxyCommand lines.
* The `GSSAPI* yes` lines do not cause problems with non-Kerberos use.
* The `PreferredAuthentications gssapi-with-mic` causes non-Kerberos ProxyJump to fail. That line removes the `password` option entirely, but it looks like the `lsst-*` machines have `gssapi-with-mic` as their first option anyway. If users have Kerberos active, it looks like it would get used anyway?

Suggestions on whether to have the Kerberos lines commented out (or just the `PreferredAuthentication` one?) with a note, or to leave them uncommented and have a note that non-Kerberos users should comment out that line? Or do we even need that?

I'll make a properly named branch/PR once the discussion here settles. I wish we could control what GitHub offers as the default name, to make that more obvious.

rra:
Yeah, setting `PreferredAuthentications` that way says to only use Kerberos. I would be hopeful that it would work for both Kerberos and non-Kerberos users if you just omitted that line. I wonder if that's the root of why the current configuration doesn't work for you.

parejkoj:
@rra : yes, that was my point with bullet 3. `PreferredAuthentications` makes this config not work for non-Kerberos users, the other options don't seem to matter. Given the default server config, I wonder if it's even necessary at all for those using Kerberos?

rra:
Right, I would hope that it's not. If you don't have Kerberos or if Kerberos fails, it should just fall back to the next mechanism, and if the servers are advertising GSSAPI first, it should try it first.

parejkoj:
Closing this, as it doesn't apply to usdf at all.

PR #484, opened 2021-05-06
Remove non-gender-neutral language

arunkannawadi:


plazas (approved):
Looks good. Thank you for this. I wonder if there are other instances in the documents that should be changed to, or if there's some way to do a systematic search. In any case, I think this is a good precedent that we should all be alert for things like these that should be changed.

arunkannawadi:
Thanks Andrés. No result shows up if you search for a specific pronoun, so it's probably not indexed and might be difficult to search systematically. It's a living document, so we should just patch them whenever we spot them.

PR #487, opened 2021-06-15
Add link to from pybind11 to python property section

parejkoj:
I noticed that the pybind11 section about python properties was still a TBD, but RFC-279 was implemented years ago. I think all we need here is a link to the python section, no?


ktlim (approved):
Yup!  Thanks for the fix.

PR #488, opened 2021-06-30
Require tickets for contributed code.

ktlim:


timj (approved):
That `git fetch origin` is pretty cool. Thanks.

The other issue I can think of is that once we create this ticket branch we are taking over the code and promising to do fixes from test failures in Jenkins and also implementing review changes and style fixes.  Will the PR run the github action to at least pass the flake8 test? Are we going to try to do a code review before we take the code in-house?

PR #491, opened 2021-07-23
Add instructions about adding package to pipelines.lsst.io

arunkannawadi:


stack/adding-a-new-package.rst
```diff
@@ -37,6 +37,8 @@ The new package then needs to be added to the :file:`ups/*.table` file (and poss
 Table files should use ``setupRequired(package_name)`` or ``setupOptional(package_name)`` as necessary; test data packages are usually optional to allow releases to be made without requiring large additional data packages to be included.
 Packages that use optional dependencies must be written to ensure that they can pass their unit tests when the package is not available.
 
+The new package may be added to the :file:`ups/pipelines_lsst_io.table` file in `lsst/pipelines_lsst_io`_ repository so that its documentation is built as HTML pages and made available at https://pipelines.lsst.io/v/daily.
```
jonathansick, line 40:
```suggestion
Finally, if the package is part of the LSST Science Pipelines, you should include it in the Pipelines documentation, https://pipelines.lsst.io.
See :doc:`/stack/add-a-package-to-pipelines-lsst-io` for instructions.
```

jonathansick (changes requested):
Thanks, this is a good idea to cross-reference here.

These steps are already documented in https://developer.lsst.io/stack/add-a-package-to-pipelines-lsst-io.html, though, so we'd rather not duplicate details about the `pipelines_lsst_io` table file across two pages. Could you please re-write the paragraph so that it links to that page to avoid duplication of content? I've included suggested content.

PR #511, opened 2021-11-28
DM-32648: Change default branch to main.

ktlim:


timj (approved):

ktlim:
Actually, @jonathansick if you could look over the change to `conf.py` only, that would be great.  I think the rest is all uncontroversial documentation.

jonathansick (approved):
Those changes should be fine.

PR #517, opened 2022-02-25
Add note about pytest log messages

parejkoj:
I keep referring back to my community post[1] on this topic, and figured it should be in the dev guide. It should maybe go into `python/testing.rst` instead, but that page says it requires an RFC for changes.

1: https://community.lsst.org/t/recent-changes-to-science-pipelines-logging/5732/4?u=parejkoj

### DIFF
--- a/stack/logging.rst
+++ b/stack/logging.rst
@@ -88,6 +88,8 @@ To specify the threshold or the lowest-severity log messages a logger handles, :
 
    logger.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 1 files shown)


ktlim:
I think I agree with you that this belongs in the `python/testing` document, perhaps in or around the `useful-pytest-options` section.  I will preemptively declare that an RFC is not necessary as this does not change any of the testing requirements, process, or mechanisms.  (And you could do this on your own as well under authority of https://developer.lsst.io/team/empowerment.html.  Furthermore, "requires an RFC" shouldn't be a barrier to putting things in the right place.)

Please move this text there.  Thanks!

PR #518, opened 2022-02-25
Add note about pytest log messages

parejkoj:
I keep referring back to my community post[1] on this topic, and figured it should be in the dev guide.

I wonder if we need to rework the paragraph just above my new one, since it refers to stdout? It also says "rare standalone usages": I don't know that that's exactly "rare"?

1: https://community.lsst.org/t/recent-changes-to-science-pipelines-logging/5732/4?u=parejkoj


parejkoj:
Here's the rendered version: https://developer.lsst.io/v/u-parejkoj-pytest_logging-1/python/testing.html#pytest-logging-output

ktlim (approved):
I approve this change.

I can believe that pytest running under scons is more prevalent than standalone pytest executions, but that may be changing.

PR #521, opened 2022-03-22
DM-33519: Update C++ style guide 5-24b for RFC-817.

TallJimbo:
Raw pointers are now preferred over shared_ptr for nullable function arguments when there is no transfer or sharing of ownership.


cpp/style.rst
```diff
@@ -1683,12 +1683,20 @@ Passing by ``const`` reference when possible is much more efficient than passing
 
 .. _style-guide-cpp-5-24b:
 
-5-24b. Smart pointers (such as ``shared_ptr``) should only be used as arguments if a reference or const reference cannot be used.
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+5-24b. Smart pointers (such as ``shared_ptr``) should only be used as arguments if a reference, const reference, or raw pointer cannot be used.
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Examples of when a smart pointer argument is appropriate include when the pointer itself may be reset, when the pointer (not the *pointee*) will be copied and held after after the function returns (as in a constructor or member function setter), or when an override of a method may need to similarly copy or hold the pointer after the function returns (even when the default base implementation does not).
```
ktlim, line 1689:
```suggestion
Examples of when a smart pointer argument is appropriate include when the pointer itself may be reset, when the pointer (not the *pointee*) will be copied and held after the function returns (as in a constructor or member function setter), or when an override of a method may need to similarly copy or hold the pointer after the function returns (even when the default base implementation does not).
```

cpp/style.rst
```diff
@@ -1683,12 +1683,20 @@ Passing by ``const`` reference when possible is much more efficient than passing
...
+In most other cases, reference or ``const`` reference arguments should be used.
+When null values are possibility, raw pointers should be used.
```
ktlim, line 1691:
```suggestion
When null values are a possibility, raw pointers should be used if the pointer is not copied.
```

ktlim (approved):
I approve these changes (with the suggested fixes).

PR #524, opened 2022-04-28
Tickets/dm 34594

womullan:
start focus times

### DIFF
--- a/work/flow.rst
+++ b/work/flow.rst
@@ -366,7 +366,8 @@ The pull request's title should be prefixed with the Jira ticket handle, followe
    DM-NNNN: {{Short summary}}
 
 This format helps you and other developers find the right pull request when browsing repositories on GitHub.
-The short summary does not need to match the Jira issue's title, but should describe the code changes.
+The short summary should describe the code changes.
+It can often be the Jira title if that is relevant to the changes.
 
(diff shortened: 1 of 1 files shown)


team/focus-times.rst
```diff
@@ -1,10 +1,20 @@
...
 
 DM will observe  **Focus Friday**. 
 
+DM will observe  **Focus Weeks** around the end/start of year and in Summer and Autumn.
+
+Focus Weeks 2022
+================
+The DM Focus weeks  will be
+- June 13-17 
+- September 18-22
```
isullivan, line 14: September 19-23?

communications/slack-culture.rst
```diff
@@ -320,7 +320,7 @@ This can be especially difficult if someone starts a thread off an older message
 Focus Friday
```
ktlim, line 320: Adjust this heading too.

frossie (changes requested):
So I obviously don't mind, but I think we had conceived of the "meeting holidays" to not be under Focus Friday rules, and instead (1) cancel all standing meetings (2) highlight to staff that it's a good time to take time off without "falling behind". This was quite a lot laxer than Focus Friday rules. We should probably have a wider discussion to make sure we agree on the specifics?

womullan:
ok on agenda for the morning

PR #532, opened 2022-06-17
Make "create a package" more prominent

parejkoj:
The "how to make a new package" instructions were buried in a hard to find paragraph in amongst the RFC stuff. It makes more sense to have it be the first thing a user sees.

As-built version:
https://developer.lsst.io/v/u-parejkoj-create-a-package/stack/adding-a-new-package.html


ktlim (approved):
Looks great.  Thanks!

PR #534, opened 2022-06-21
Update list of testdata packages in shared path

parejkoj:


ktlim (approved):
Maybe alphabetize the list?

I wonder if the cron job logs should be put in a subdirectory instead of clogging up the main directory.

ktlim:
I moved the logs and changed the cron job to put them in a subdirectory.

PR #535, opened 2022-06-28
DM-34974: dev guide updates in preparation for bootcamp

parejkoj:


team/onboarding.rst
```diff
@@ -80,6 +80,8 @@ Ask your T/CAM for access to the ‘lsstc’ Slack team.
 You can access it online at https://lsstc.slack.com, and through `Slack's mobile and desktop apps <https://get.slack.help/hc/en-us/articles/201746897-Slack-apps-for-computers-phones-tablets>`__.
 Slack's `online help <https://get.slack.help/hc/en-us>`__ is a great way to learn Slack's features.
 
+Be sure to :doc:`link your github profile to your slack account <../communications/slack-github-username>` so that our slack integrations will work correctly.
```
ktlim, line 83: You should use "GitHub" and "Slack" capitalizations throughout.

ktlim (approved):

git/git-lfs.rst
```diff
@@ -49,15 +49,12 @@ Configuring Git LFS
 Basic configuration
 -------------------
 
-Once, after you've installed Git LFS, run:
+Run this command to add a ``filter "lfs"`` section to :file:`~/.gitconfig`:
```
ktlim, line 52: I think it's still good to say that this only needs to be done once per user (home directory), but it does need to be done before using Git LFS.

work/flow.rst
```diff
@@ -284,7 +282,7 @@ Before offering unsolicited code review on your colleagues' user branches, remem
 Developers can feel free to rebase and force push work to their personal user branches.
 
 A user branch *cannot* be merged into main; it must be converted into a *ticket branch* first.
-The only exception is for documentation edits and additions.
+The only exception is for documentation fixes.
```
ktlim, line 285: I'm not sure that everyone would interpret "fixes" to include "additions".

work/flow.rst
```diff
@@ -295,7 +293,7 @@ Ticket branches are associated with a JIRA ticket.
 Only ticket branches can be merged into ``main``.
 (In other words, developing on a ticket branch is the only way to record earned value for code development.)
 
-If the JIRA ticket is named ``DM-NNNN``, then the ticket branch will be named
+If the JIRA ticket is named ``DM-NNNN``, then the ticket branch MUST be named
```
ktlim, line 296: Maybe add an extra "N" since we're well beyond 10000?

ktlim (approved):
A few quibbles with this batch, but overall looks good.

work/flow.rst
parejkoj: My assumption was that additions don't count, because they can break things. Maybe this, which I think captures the intent?

```
The only exception is for documentation fixes in packages, or edits and additions to this developer guide.
```

ktlim: The original intent was to try to make contributing to the documentation as low-cost as possible, even removing the slight friction of creating a Jira ticket describing what documentation was to be added or modified.  That includes docstrings, package doc pages, or guides.

Even PRs from `u` branches still get GHA checks and merge protections.

parejkoj: Thanks for clarifying: I've made use of that. How about this version?

https://github.com/lsst-dm/dm_dev_guide/pull/535/commits/34ab153a94f69d1d2818d199f81be0b275ac98b0

ktlim: Looks great.

PR #536, opened 2022-06-28
DM-34974: Add merge vs. rebase image and paragraph

parejkoj:


work/flow.rst
```diff
@@ -503,7 +503,18 @@ If a rebase was required, a final check with Jenkins should be done.
 We encourage you to use this button when GitHub says "This branch has no conflicts with the base branch", "All checks have passed", and at least one of the checks has "Required" next to it, which should be the case for almost all repos that are being changed.
 For normally-configured repos, using the button will also delete the ticket branch after the merge.
 Do *not* select a different merge strategy from the pulldown next to the button; these should be disabled anyway.
-Also do *not* use the "Update my branch" button as that does a merge; rebase from the command line instead.
+Also do *not* use the "Update my branch" button as that does a merge from main; rebase from the command line instead.
+
+The figure below compares the git history repository that uses a "merge from main" workflow vs. our "rebase from main" workflow.
```
timj, line 508: I think the left picture is a "merge when approved as-is" workflow rather than a merge from main to branch workflow (which would look even worse because it will have an extra merge commit before the merge back to main). The left is a workflow where you create a branch and work on it for a while and then you merge when you are done and then I think will include conflict resolution as part of the merge commit itself. Our merge commits are empty. I think GitHub only gives you the update-my-branch button if there are no conflicts (but I don't use it so I'm not sure).

parejkoj: Hmmm... Is it worth keeping something like this in the dev guide? Should we just show the right side? I don't know that it's worth trying to redo it, unless you can make each of them quickly: my preferred git gui doesn't show this kind of graph in a way that would work well for an image.

timj: I'm fine with just showing the right hand side and making that point that it's much easier to see what changed and git bisect is more obvious. That's what I did in the SPIE paper describing our process.

timj: For my talk at SPIE I did show the left hand side but not for the paper itself.

timj (approved):

PR #537, opened 2022-06-28
DM-34974: add note about where to put ndarray types

parejkoj:


parejkoj:
@ktlim : changes to our numpydoc guide requires System Architect approval. I think this change wouldn't need an RFC, but please let me know.

@nevencaplar : is this change helpful in answering your question from Slack? Here's what the built version looks like: https://developer.lsst.io/v/tickets-DM-34974-ndarray/python/numpydoc.html#array-types

nevencaplar:
Yes, I think that is a helpful change.

nevencaplar (approved):
Yes, I think that is a helpful change.

ktlim (approved):
LGTM.

PR #541, opened 2022-08-02
DM-35567: Add guide on topic-based writing for users

jonathansick:
This page is based on talks I've given to DM on organizing documentation around topics and using topics types to create consistent documentation structures.

See draft: https://developer.lsst.io/v/DM-35567/user-docs/topic-based-documentation.html

Question for review — this page has initially involved into two themes: exploring topic-based technical writing, and then designing content for users (i.e. headers, short paragraphs, lists, code samples). I'm wondering whether this page works well as it is, or if it should be split up?

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -473,6 +473,7 @@ User documentation style
 
 - :doc:`user-docs/index`
 - :doc:`user-docs/lsst-specific-content-style-guide`
+- :doc:`user-docs/topic-based-documentation`
 
 .. _part-services:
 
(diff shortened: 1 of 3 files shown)


rra (approved):
This looks great to me.

I considered whether it should be split up and a few different options for splitting it up, but while it does feel a bit long, I think any way of splitting it would have to move the bulk of the page (what topic-based writing means) into a single page, and the remaining content is too short for a single page, IMO. So I'd leave it all together.

PR #551, opened 2022-08-24
DM-35388: Fix dev guide links in "tune your shell for git"

parejkoj:
The rendered version: https://developer.lsst.io/v/DM-35388/git/setup.html#tune-your-shell-for-git

### DIFF
--- a/git/setup.rst
+++ b/git/setup.rst
@@ -132,8 +132,8 @@ Tune your shell for Git
 You can build an effective development environment and workflow by tuning your Git setup.
 Here are some ideas:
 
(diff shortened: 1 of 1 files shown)


bsmartradio (approved):
Comments are helpful at explaining how to install the changes.

erinleighh (changes requested):
Links go where they say they're going to go and the expanded descriptions are helpful, however the numbers are off (skipped 2. and went from 1. to 3.) and is no longer automatically formatting as a list.

parejkoj:
Good catch on the skipped number and broken list, @erinleighh . I made a tweak after @bsmartradio looked at it, and apparently broke it then. Should be fixed now.

erinleighh (approved):
Fix looks great!

PR #560, opened 2022-09-30
DM-36409: Update stack-os-matrix docs to reflect current system

kfindeisen:
This PR updates the description of the `stack-os-matrix` job to reflect current practice, and adds some notes about gotchas that I've run into myself.

The built documentation can be found at https://developer.lsst.io/v/DM-36409/stack/jenkins-stack-os-matrix.html.

### DIFF
--- a/stack/jenkins-stack-os-matrix.rst
+++ b/stack/jenkins-stack-os-matrix.rst
@@ -34,6 +34,10 @@ Running a stack-os-matrix job
      - ``ci_hsc`` exercises most of the Data Release Production pipelines, including single-frame processing and coaddition, on HSC engineering data. This package is also run as part of the nightly build.
      - ``ci_imsim`` runs the same on simulated Rubin Observatory data. Like ``ci_hsc``, it's run nightly in addition to as part of ``stack-os-matrix``.
 
+   - **Set the conda environment** to run on.
+     The default is the latest ``rubin-env`` package.
+     You should only need to change this when testing a new environment, or to temporarily work around bugs introduced in the latest version.
+
 4. Click the **Run** button in the dialog to start the build.
 
 Monitoring the run status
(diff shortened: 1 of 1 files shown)


stack/jenkins-stack-os-matrix.rst
```diff
@@ -20,18 +20,25 @@ Running a stack-os-matrix job
...
 
    - **Set the list of EUPS packages** to build.
-     Use the default (``lsst_distrib``) to build and test your changes with a full Stack.
-     To improve build times you can instead specify the name of the package you are actively developing and check the **Skip Demo** box.
+     Use the default (``lsst_distrib lsst_ci``) to build and test your changes with a full Stack.
+     To improve build times you can instead specify the name of the package you are actively developing.
+     Before you merge a ticket branch to ``main``, **you must** run a stack-os-matrix Job with at least the default packages so that the full Stack is built and tested with your changes.
 
-   - **Check the Skip Demo** option only if you are testing your package alone (in conjunction with specifying your package in the prior text box).
+     There are also several packages that can be appended to the default to do more thorough testing at the cost of much longer build times:
```
ktlim, line 33: You might also note that ci_hsc and ci_imsim will be run automatically on `main` each night.

timj (approved):
Looks okay. It would be nice if we can eventually rename ci_hsc_gen3 and ci_cpp_gen3 so we don't have to have stub packages with the shortened name but that's not relevant for this PR.

stack/jenkins-stack-os-matrix.rst
```diff
@@ -20,18 +20,25 @@ Running a stack-os-matrix job
...
 
-     Before you merge a ticket branch to ``main``, **you must** run a stack-os-matrix Job without **Skip Demo** enabled so that the full Stack is built and tested with your changes.
+     - ``ci_cpp`` exercises the Calibration Products Pipeline in ways that are too computationally expensive for unit tests in ``cp_pipe``.
+     - ``ci_hsc`` exercises most of the Data Release Production pipelines, including single-frame processing and coaddition, on HSC engineering data.
```
timj, line 36: Late comment -- do we want to give guidance for when someone has to do this? Do science pipelines want at least one of ci_hsc or ci_imsim to be run for every single change to a pipelines package?

kfindeisen: My intent in describing what the packages do was that that would _be_ the guidance.

I would oppose a rule that broad on the grounds that it makes things unnecessarily hard to merge. Personally, I only run the big packages when I have a change that I'd reasonably expect to affect the pipelines (e.g., changes to `Exposure`, but not `ap_verify`).

In any case, we have never had such a policy before, so I think adopting it would require an RFC.

PR #577, opened 2022-12-09
Add object store information to USDF section.

ktlim:


usdf/object-store.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Prompt Processing development uses a pair of buckets to act as its "central store" and raw image storage (simulating the embargo production storage) as well as another bucket to hold test data.
+
+- rubin-pp
+- rubin-pp-users
+- rubin-prompt-processing-test
+
+Alert Production development uses a bucket to hold cutout visualizations of DiaSources for algorithm analysis.
+
+- rubin-ap-cutouts
```
parejkoj, line 43: I don't know if we want to mention this here: it should not be used as a reliable source of cutouts (I plan to purge it semi-regularly, as we generate new datasets), and I'm not sure I want people to assume that things in it will exist for any length of time. If that's ok, then it can stay.

usdf/object-store.rst
```diff
@@ -0,0 +1,86 @@
...
+
+
+Credentials
+===========
+
+A single set of credentials is used for read-only access to the raw data buckets and for read-write access to the ``-users`` buckets.
+This set can most easily be accessed by logging into the USDF RSP and starting a notebook server.
```
parejkoj, line 50: I'd put this line at the top: "The default set of credentials for read-only access is created the first time you log in to the USDF RSP and start a notebook server.", and then reword the stuff below a bit.

usdf/object-store.rst
```diff
@@ -0,0 +1,86 @@
...
+This will create a ``~/.lsst/aws-credentials.ini`` file with the credentials set as the default profile.
+The RSP and the default scripts executed by .bashrc upon ssh login will set the ``AWS_SHARED_CREDENTIALS_FILE`` environment variable to point to this file.
+
+Additional non-default profiles can be added to ``~/.lsst/aws-credentials.ini``, but note that they will be overwritten on the next USDF RSP notebook startup.
```
parejkoj, line 54: Oh! I didn't know that. That means that people absolutely shouldn't use this file for credentials management. "it may be better" below needs to become "Users should not use aws-credentials.ini and instead should..."

parejkoj (commented):
Thank you, this is a good start. If things get overwritten, we don't want users to touch them.

PR #582, opened 2023-01-04
DM-37425: Update documentation for doc/conf.py

jonathansick:
- The template already sets the package name, so no need to reference the project_name variable or version (which is no longer settable).
- Clarify that the file shouldn't be modified.


parejkoj (approved):
Thanks, this is clearer!

PR #594, opened 2023-02-17
Document workaround for git lfs on small memory.

ktlim:
This problem was discovered by Marcella Di Criscienzo in https://lsstc.slack.com/archives/CAS7Y9ADS/p1676545637893539


timj (approved):
Interesting

PR #600, opened 2023-03-28
DM-36947: Make it clear that downstream usage should be fixed when deprecating.

TallJimbo:
This includes a recipe for finding downstream usage via the PYTHONWARNINGS environment variable, which will only work when the sconsUtils branch for this ticket is also merged to main (since at present SCons does not forward that variable to tests).

This implements RFC-881 and RFC-659.


stack/deprecating-interfaces.rst
```diff
@@ -167,3 +171,23 @@ After deprecating a package as described :ref:`above <package-deprecation>`, the
...
+
+For all Python deprecations (including pybind11 and config deprecations), developers should find and fix downstream usage of a deprecated interface by turning the new warnings into errors temporarily, and running Jenkins (or running lsstsw locally).
+The easiest approach is to modify the ``PYTHONWARNINGS`` environment variable in the EUPS table file of the package containing the deprecated interface, e.g.
+
+.. code::
+
+    envPrepend(PYTHONWARNINGS, "error::FutureWarning:lsst.daf.butler.core.dimensions._universe", ",")
+    envPrepend(PYTHONWARNINGS, "error::FutureWarning:lsst.daf.butler.core.dimensions._coordinate", ",")
+
+Each module with a new deprecation must be listed separately and fully-qualified; including the module is usually sufficient for matching only the desired warnings, and is much less error-prone that including the message.
```
ktlim, line 188:
```suggestion
Each module with a new deprecation must be listed separately and fully-qualified; including the module is usually sufficient for matching only the desired warnings, and is much less error-prone than including the message.
```

PR #603, opened 2023-04-21
Update USDF batch description.

ktlim:


jchiang87 (approved):
These changes look good to me.  However, I've been wondering if we should give guidance on choices of node count vs # cpus, e.g., if one wants 128 cores, either `-n 128 -c 1` or `-n 4 -c 32` would provide that, but maybe one is better than the other in some circumstances?  I've been doing the latter, but the former seems more responsive to changes in the number of cores needed by the workflow at any given time.

ktlim:
I will ask Greg Daues for advice on `-n`/`-c` on Monday.

PR #604, opened 2023-04-24
Add suggestions for allocate_nodes options.

ktlim:


daues (approved):
Approving the current draft as it is important to guide users away from running a few hundred glide-in's to get a few hundred cores, and so the mental picture of n=[1,32]  achieves that. We'll try to think of a good way to indicate that n=32 is not actually a scheduler or scaling limit for larger scale work.

PR #606, opened 2023-04-25
DM-38873: Remove outdated instructions

bsmartradio:
I removed what I think are all the outdated instructions, but please let me know if I need to keep some of the removed text in, change anything else. I wanted to get this moved as soon as possible since others have also run into the issue of their stack breaking when following the current instructions and it can be frustrating. If you'd prefer it left alone for now I can close the pull request.


jonathansick (approved):

jonathansick:
Looks good, thanks.

PR #610, opened 2023-05-09
Clarification for user branches

bsmartradio:
Added clarification that it should be your github username on user branches, not usdf or other LSST/Rubin usernames.

If this should be left as is, let me know and I'll close the PR.

### DIFF
--- a/work/flow.rst
+++ b/work/flow.rst
@@ -274,7 +274,7 @@ These branches are named
 
 .. code-block:: text
 
-   u/{{username}}/{{topic}}
+   u/{{github-username}}/{{topic}}
 
 User branches can be pushed to GitHub to enable collaboration and communication.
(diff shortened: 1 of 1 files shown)


ktlim:
Sorry I didn't see this.  (It's best not to rely on GitHub email notifications for reviews; a separate Slack ping helps if it's not a Jira ticket that's being reviewed.)

The main purpose of having user branches is so that users don't collide with each other and so that they don't collide with ticket branches.  Otherwise I don't see a need for automation that would parse the ticket branch name.  So I think whatever (unique) username the user is comfortable with should be acceptable.

Admittedly I have a stake in this as I have many `u/ktl` branches and (virtually?) no `u/ktlim` ones.

bsmartradio:
> 

Ah alright, that makes sense. I'll close the PR then.

PR #613, opened 2023-05-15
DM-39199: Add a note about FITS libraries from Python

arunkannawadi:
as requested by the DM-CCB on RFC-921.

### DIFF
--- a/python/astropy.rst
+++ b/python/astropy.rst
@@ -13,10 +13,6 @@ Integration of Astropy core into the LSST software stack is an ongoing process t
 This document is not discussing Astropy affiliated packages, use of which must go through the standard :doc:`RFC </communications/rfc>` process.
 Not all Astropy core packages can be used by default.
 
-The following packages should be used from Astropy in preference to other Python packages:
-
-* :mod:`astropy.io.fits` instead of ``pyfits`` (use of ``pyfits`` is now deprecated).
-
 The following packages can be used internally in packages if they do not leak into public APIs:
 
 * :mod:`astropy.coordinates`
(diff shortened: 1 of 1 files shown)


python/astropy.rst
```diff
@@ -28,6 +24,15 @@ The following packages can be used internally in packages if they do not leak in
 * :mod:`astropy.cosmology`
 * :mod:`astropy.visualization`
 
+For reading and writing files in FITS format, both :mod:`astropy.io.fits` (use of ``pyfits`` is now deprecated) and `fitsio`_ are allowed.
```
timj, line 27: I'm guessing at this point we can drop the ``pyfits`` comment since it's been dead for years and many people don't even know it existed.

arunkannawadi: Am I old enough to have actually used it? 😢

python/astropy.rst
```diff
@@ -28,6 +24,15 @@ The following packages can be used internally in packages if they do not leak in
...
+While the latter usually surpassess the former in performance, developers must explicitly test with both these libraries and make a choice based on what works best for their use case.
```
ktlim, line 28: "surpasses".

I understand that there are differences in the functionality available between the two, which means that I'd rather say "developers must evaluate which works best for their use case" without requiring explicit testing.

ktlim (approved):

arunkannawadi:
Typo fixed, comment about `pyfits` removed and wording changed slightly.

python/astropy.rst
```diff
@@ -28,6 +24,16 @@ The following packages can be used internally in packages if they do not leak in
 * :mod:`astropy.cosmology`
 * :mod:`astropy.visualization`
 
+For reading and writing files in FITS format, both :mod:`astropy.io.fits` and `fitsio`_ are allowed.
+These libraries differ in the features available and in performance (with the latter usually surpassing the former).
+Developers must evaluate which works best for their use case and choose one accordingly..
```
ktlim, line 29:
```suggestion
Developers must evaluate which works best for their use case and choose one accordingly.
```

PR #616, opened 2023-05-25
Update git credential and auth token text

parejkoj:


git/setup.rst
```diff
@@ -111,18 +111,44 @@ To have your credentials cached for 1 hour (3600 seconds):
...
+Due to GitHub's authentication interface, you must use a personal access token instead of a password when pushing to GitHub via HTTPS.
+GitHub access via ssh does not require a token, but :doc:`Git LFS repos <git-lfs>` work best via HTTPS.
+A credential helper, as :ref:`described above <git-credential-helper>`, will let you cache this token, so you do not have to save it (you can only view it on GitHub when you generate it).
+Create a personal token at `GitHub Personal Access Tokens <https://github.com/settings/tokens>`_, using the "generate new token" button in the upper right, with the "classic" interface.
+Give it a name (e.g. "LSST USDF"), choose an expiration date (when the token expires you will have to generate a new one and re-enter it into your credential cache), and set the following permissions::
+
+  repo->public_repo
+  admin:org->read:org
+
+!!! NOTE: do we also need ``workflow``, to be able to change GitHubActions, or any others? Should we have a note about other options that people may want?
```
timj, line 138: Yes, it makes sense for people to include that in their token if they are allowed.

git/setup.rst
```diff
@@ -111,18 +111,44 @@ To have your credentials cached for 1 hour (3600 seconds):
...
+   Username for 'https://github.com': <GitHub username>
+   Password for 'https://<user>@github.com': <GitHub token>
+   Username for 'https://git-lfs.lsst.codes': <GitHub username>
+   Password for 'https://<user>@git-lfs.lsst.codes': <GitHub token>
+
+At the prompts, enter your GitHub username and token.
+
+Once your credentials are cached, you won't need to repeat this process on your system (:ref:`unless you opted for the cache-based credential helper <git-credential-helper>`) until the token expires.
+
+If you find that ``git push`` is not working but also not asking you for credentials, you may need to manually insert the username and token into the credential store or macOS keychain. NOTE: we should expand on this, as I recall having this problem but don't know what I did to fix it.
```
timj, line 151: On macOS I think I had to run up Keychain Access and find the token and delete it so it would prompt me again.

timj (approved):
Seems fine. I will defer to @ktlim though.

git/setup.rst
```diff
@@ -111,18 +111,44 @@ To have your credentials cached for 1 hour (3600 seconds):
...
+-------------------------
+
+Due to GitHub's authentication interface, you must use a personal access token instead of a password when pushing to GitHub via HTTPS.
+GitHub access via ssh does not require a token, but :doc:`Git LFS repos <git-lfs>` work best via HTTPS.
+A credential helper, as :ref:`described above <git-credential-helper>`, will let you cache this token, so you do not have to save it (you can only view it on GitHub when you generate it).
+Create a personal token at `GitHub Personal Access Tokens <https://github.com/settings/tokens>`_, using the "generate new token" button in the upper right, with the "classic" interface.
+Give it a name (e.g. "LSST USDF"), choose an expiration date (when the token expires you will have to generate a new one and re-enter it into your credential cache), and set the following permissions::
+
+  repo->public_repo
+  admin:org->read:org
```
ktlim, line 136: This is misleading.  `admin:org` should not be checked.  `repo` may or may not need to be checked; mostly we avoid private repos, but there are a few.

git/setup.rst
ktlim: It is rare for people to change GHAs, especially for git lfs repos, but it might be weird if it failed due to token scope.

ktlim (approved):
Mostly looks good.

One should generally _not_ grant `admin:org` access.

The final NOTE needs to be cleaned up.

git/setup.rst
```diff
@@ -111,18 +111,44 @@ To have your credentials cached for 1 hour (3600 seconds):
...
 
-If you find that ``git push`` is not working but also not asking you for credentials, you may need to manually insert the username/password or token into the credential store or macOS keychain.
+.. _git-auth-tokens:
 
-The DM Git LFS documentation has further information about :ref:`authenticating with our LFS storage backend <git-lfs-auth>`.
+Git authentication tokens
+-------------------------
+
+Due to GitHub's authentication interface, you must use a personal access token instead of a password when pushing to GitHub via HTTPS.
+GitHub access via ssh does not require a token, but :doc:`Git LFS repos <git-lfs>` work best via HTTPS.
```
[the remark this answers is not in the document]
parejkoj: It was definitely true in the past; lets leave it as-is for now, until we've had a chance to test it better.

PR #618, opened 2023-05-30
Add searching for TODO recommended

arunkannawadi:
This patch makes searching for TODO comments with current ticket number an explicit step in the development workflow.

### DIFF
--- a/work/flow.rst
+++ b/work/flow.rst
@@ -323,6 +323,16 @@ Review Preparation
 When development on your ticket branch is complete, we use a standard process for reviewing and merging your work.
 This section describes how to prepare your work for review.
 
+.. _workflow-search-todo:
+
+Search for TODO comments
+------------------------
+
+Search for your ticket number (with the 'DM-' prefix) in all repositories associated with the LSST organization on GitHub (using "org:lsst" as the prefix).
+If there are any TODO comments linked with your current ticket, make sure that they are addressed in this ticket and remove the comments.
+If any additional follow up work is needed to fully resolve the issue, update the comment to reflect the follow up ticket.
+This is particularly relevant for long-standing tickets.
+
 .. _workflow-pushing:
 
 Pushing code


arunkannawadi:
@ktlim would you be able to review this?

work/flow.rst
```diff
@@ -511,7 +520,6 @@ This history makes ``git bisect`` searches more obvious.
 We understand that this workflow may take longer to get used to, but the end result of a clean history has paid us dividends in finding when particular algorithmic changes had unexpected effects.
 
 .. figure:: /_static/development/git-rebase-history.png
-   :target: ../_images/git-rebase-history.png
```
ktlim, line 514: What's wrong with keeping this?  It allows people to expand the image if desired.

work/flow.rst
```diff
@@ -323,6 +323,15 @@ Review Preparation
 When development on your ticket branch is complete, we use a standard process for reviewing and merging your work.
 This section describes how to prepare your work for review.
 
+.. _workflow-search-todo:
+
+Search for TODO comments
+------------------------
+
+Search for your ticket number in all repositories associated with the LSST organization on Github (using "org:lsst" as the prefix).
```
ktlim, line 331: This seems like a reasonably quick and useful step.  Are you confident that there will not be many false positive hits?  Are you confident that this is uncontroversial enough that we don't need an RFC for it?  How are you planning on announcing this?

arunkannawadi: I occasionally do search for `DM-NNNNN` and I rarely see any false positives (just searching for the number does give a lot of false positives though). May be I should clarify to include the `DM-` prefix. 

I think an announcement on the #dm channel would suffice. I wasn't thinking that this would require RFC, but can do one if you think it'll get it noticed more.

I do think it's non-controversial since it's a quick step and it is applicable only when a ticket has been lingering around for a long time. That's something I should add as well.

work/flow.rst
arunkannawadi: This was an attempt to fix the broken linkcheck step in the Weekly CI. The more recent runs seem to be broken on other things, making me think that this is flaky. I'll exclude this commit from the PR.

https://github.com/lsst-dm/dm_dev_guide/actions/runs/5239410991/jobs/9459211675

ktlim (approved):

PR #620, opened 2023-06-02
DM-38916: Document edge-case behavior of Python type descriptions

kfindeisen:
This PR adds a couple of notes to the documentation clarifying how to do multi-line PRs and how to resolve relative type names.

Compiled version at https://developer.lsst.io/v/DM-38916/python/numpydoc.html.

### DIFF
--- a/python/numpydoc.rst
+++ b/python/numpydoc.rst
@@ -643,6 +643,14 @@ If the type of the sequence's items isn't known, leave out the item typing and e
    items : `list`
        A sequence of items, which can be any type.
 
+Sequences of complex types can potentially have very long type descriptions.
+If such a description must be split across multiple lines, you must end each line but the last with a ``\``:
+
+.. code-block:: rst
+
+    items : `lsst.afw.collections.FancyList` [`~collections.abc.Mapping` \
+            [`str`, `~lsst.pex.config.Config`]]
+
 .. _py-docstring-parameter-types-dict:
 
 Dictionary types
@@ -705,6 +713,14 @@ If the keys are well-known, document the keys using a pattern similar to :ref:`p
            all content is aligned with the opening content line.
 
 Notice how the keys are shown by enclosing the quoted strings in double backticks to clarify that the keys are `str` types.
+
+Mappings of complex types can potentially have very long type descriptions.
+If such a description must be split across multiple lines, you must end each line but the last with a ``\``:
+
+.. code-block:: rst
+
+    items : `lsst.afw.collections.FancyList` [`~collections.abc.Mapping` \
+            [`str`, `~lsst.pex.config.Config`]]
 
 .. _py-docstring-parameter-types-struct:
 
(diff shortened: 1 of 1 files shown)


python/numpydoc.rst
```diff
@@ -636,6 +636,11 @@ The type inside the brackets is the type of each item.
 
    Leaving out the space results in a parsing error.
 
+.. warning::
+
+   The parameter name and type description must fit on a single line.
```
timj, line 641: This is not correct. As you can see in lsst/daf_butler#840 you can escape the line with a backslash. I only realized this last week when I saw @TallJimbo do it in some code.

kfindeisen: Confirmed; I'll do a few more experiments and document the behavior.

python/numpydoc.rst
```diff
@@ -589,9 +589,15 @@ This works for both built-in types and most importable objects:
        [...]
 
 In general, provide the full namespace to the object, such as ```lsst.pipe.base.Struct```.
-It may be possible to reference objects in the same namespace as the current module without any namespace prefix.
+It may be possible to reference objects in the same package as the current type or function without any namespace prefix.
 Always check the compiled documentation site to ensure the link worked.
 
+.. note::
+
+   Relative links to types are evaluated relative to the *package* in which the linking object is documented, not the module file in which the docstring is written.
```
timj, line 597: I also saw in lsst/daf_butler#846 that a `.` prefix can be used in numpydoc to allow relative to current.

python/numpydoc.rst
```diff
@@ -589,9 +589,15 @@ This works for both built-in types and most importable objects:
...
+   In particular, importing a type into a module does not let you use unqualified links to that type from module code.
```
timj, line 598: Sphinx gets type annotations correct in the signature these days even if it doesn't get the numpydoc part correct. I'm not sure where we should be mentioning type annotations.

How does something like a bare `DatasetType` work in numpydoc then if we say that importing into a module does not let you use it? lsst/daf_butler#846 does do that sometimes.

python/numpydoc.rst
```diff
@@ -706,6 +720,14 @@ If the keys are well-known, document the keys using a pattern similar to :ref:`p
 
 Notice how the keys are shown by enclosing the quoted strings in double backticks to clarify that the keys are `str` types.
 
+Mappings of complex types can potentially have very long type descriptions.
```
timj, line 723: I guess the duplication is okay since people will potentially be coming at this from different directions.

timj (approved):
Looks like a nice cleanup although I have questions about the handling of relative links.

python/numpydoc.rst
kfindeisen: Both [`Registry`](https://pipelines.lsst.io/v/daily/py-api/lsst.daf.butler.Registry.html) and [`DatasetType`](https://pipelines.lsst.io/v/daily/py-api/lsst.daf.butler.DatasetType.html) are documented in the same package, `lsst.daf.butler`. Their modules are irrelevant.

PR #621, opened 2023-06-16
Add note about macOS git completion to git setup

parejkoj:


timj (approved):
Thanks. Didn't know that.

PR #623, opened 2023-06-27
DM-39713: Describe ci_middleware and clarify roles of various ci_* packages.

TallJimbo:


parejkoj (approved):
Looks good. My only question is whether we should mention that people might want to submit separate Jenkins jobs for each ci package they're testing, so that the whole build doesn't take 4-6 hours.

TallJimbo:
I tend to launch everything in one Jenkins job because I worry about monopolizing too many of the workers at once, but I don't know if that's a real concern.

parejkoj:
I've had jobs timeout when running ci_hsc+ci_imsim, and Tim has suggested submitting multiple jobs. Might as well just merge this udpate and we can ask on #dm-arch if they have a preference and tweak it later?

PR #624, opened 2023-06-28
DM-24949: Clarify dev guide documentation on adding a new package

enourbakhsh:


stack/adding-a-new-package.rst
```diff
@@ -15,6 +15,10 @@ DM packaging of third party code should proceed as described in :doc:`packaging-
 
 If the new package needs a distinct Jira component (most will), any DMLT member (such as your manager) can add one.
 
+Upon adding a new package, the dependency should be listed in the table file of packages that specifically utilize it. This method prevents the code from breaking for users who only setup specific packages they are utilizing and not the entire ``lsst_apps`` or ``lsst_distrib``. Even when changes are made on a branch, Jenkins will promptly detect the integration of the new package and its dependencies as soon as it picks up the ``ap_association`` change, and proceed to pull in the new package into the build. Dependencies managed through Conda, rather than EUPS, are assumed and do not need to be specified in the table file.
```
timj, line 18:
```suggestion
Upon adding a new package, the dependency should be listed in the table file of packages that specifically utilize it. This method prevents the code from breaking for users who only setup specific packages they are utilizing and not the entire ``lsst_apps`` or ``lsst_distrib``. Even when changes are made on a branch, Jenkins will promptly detect the integration of the new package and its dependencies as soon as it picks up the ``ap_association`` change, and proceed to pull in the new package into the build. Dependencies managed through Conda, rather than EUPS, are assumed and should not be specified in the table file.
```

If you specify a conda package in your table file eups will get upset. I've changed the emphasis to indicate this. If you have a C++ linking dependency you have to specify that in the .cfg file.

ktlim, line 18:
```suggestion
Upon adding a new package, the dependency should be listed in the table file of packages that specifically utilize it. This method prevents the code from breaking for users who only setup specific packages they are utilizing and not the entire ``lsst_apps`` or ``lsst_distrib``. Even when changes are made on a branch, Jenkins builds of that branch will use the new table file and will proceed to pull the new package into the build. Dependencies managed through Conda, rather than EUPS, are assumed and do not need to be specified in the table file.
```

ktlim (approved):
Thanks for adding this!

stack/adding-a-new-package.rst
timj: @ktlim  where do you stand on my comment?

ktlim: Oh, I fully agree, and that change should be included as well.

PR #630, opened 2023-07-12
DM-39845: Add a section on ruff configuration

timj:


python/style.rst
```diff
@@ -151,6 +151,66 @@ The ``exclude`` field lists paths that are not usefully linted by :command:`flak
 Auto-generated Python should not be linted (including :file:`bin/` for Stack packages with :file:`bin.src/` directories).
 We also discourage linting :file:`__init__.py` modules due to the abundance of :pep:`8` exceptions typically involved.
 
+.. _style-guide-py-ruff:
+
+Code MAY be validated with ruff
+-------------------------------
+
+The ruff_ tool may be used to validated Python source code against the portion of :pep:`8` adopted by Data Management.
```
ktlim, line 159:
```suggestion
The ruff_ tool may be used to validate Python source code against the portion of :pep:`8` adopted by Data Management.
```

Also, why/when `ruff_` here and `:command:ruff` later?

python/style.rst
```diff
@@ -151,6 +151,66 @@ The ``exclude`` field lists paths that are not usefully linted by :command:`flak
...
+ruff_ is significantly faster than :command:`flake8` to execute and this can be useful for larger projects.
+It also is far more configurable with many built in linting extensions.
```
ktlim, line 161:
```suggestion
It also is far more configurable with many built-in linting extensions.
```

python/style.rst
```diff
@@ -151,6 +151,66 @@ The ``exclude`` field lists paths that are not usefully linted by :command:`flak
...
+Sometimes ruff_ will suggest fixes that can be automated.
+Use the ``--fix`` option to apply the automated fixes and then examine them before committing them to make sure they look correct.
```
ktlim, line 163:
```suggestion
Use the ``--fix`` option to apply the automated fixes, and then examine them before committing them to make sure they look correct.
```

This is also redundant with the text at the end of the configuration section.  Please choose one or the other.

python/style.rst
```diff
@@ -151,6 +151,66 @@ The ``exclude`` field lists paths that are not usefully linted by :command:`flak
...
+
+.. _ruff: https://beta.ruff.rs/docs/
+
+.. _style-guide-py-ruff-install:
+
+Ruff installation
+^^^^^^^^^^^^^^^^^
+
+Ruff comes installed with any `LSST Science Pipelines installation <https://pipelines.lsst.io/#installation>`_ using ``rubin-env`` 7.0 or newer.
+It is run automatically when building packages with ``scons`` if a :command:`ruff`` configuration is found.
```
ktlim, line 173:
```suggestion
It is run automatically when building packages with ``scons`` if a :command:`ruff` configuration is found.
```

python/style.rst
```diff
@@ -151,6 +151,66 @@ The ``exclude`` field lists paths that are not usefully linted by :command:`flak
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Science Pipelines packages do not come with a :command:`ruff` configuration by default and one must be added if you wish to use the linting tool with your package.
+Once configured SCons will prefer to run this tool over :command:`flake8`.
+The configuration for :command:`ruff` must be added to the ``pyproject.toml`` file -- this will be the same file used to configure other tools such as :ref:`black and isort<formatting-python-code>`.
+
+.. literalinclude:: examples/pyproject-ruff.toml
+  :language: toml
+
+There is no need to explicitly specify ``W503`` since :command:`ruff` assumes that by default.
```
ktlim, line 187: I think you need to say that this configuration is 1) mandatory and 2) sufficient if used at all, in addition to the optional parts mentioned later.  In other words, you're not allowed to leave parts out as that would reduce the conformance checking, and there's nothing else that you need to watch out for separately that is not caught by ruff.

ktlim (approved):

python/style.rst
timj: The elephant in the room is https://github.com/astral-sh/ruff/pull/4694 which is why we do currently need flake8 for full compliance.

PR #632, opened 2023-07-18
Specify that __all__ should be a tuple

parejkoj:
We've had some conversation about this on slack in the past, and settled on tuple, but never specified it in the dev guide.

### DIFF
--- a/pybind11/how-to.rst
+++ b/pybind11/how-to.rst
@@ -960,7 +960,7 @@ We'll use the ``continueClass`` decorator to reopen the class and add a new meth
 
     from ._tmpl import ExampleTwo
 
-    __all__ = []  # import for side effects
+    __all__ = ()  # import for side effects
 
 
     @continueClass
@@ -999,7 +999,7 @@ As with ``ExampleTwo``, add this to a pure-Python ``_ExampleThree.py``:
 
     from ._tmpl import ExampleThreeI, ExampleThreeD
 
-    __all__ = ["ExampleThree"]
+    __all__ = ("ExampleThree", )
 
 
     class ExampleThree(metaclass=TemplateMeta):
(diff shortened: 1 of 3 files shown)


timj:
A tuple seems right to me given that you don't want someone to change the content, but there is a problem that the Python docs themselves use a list: https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

Not sure if we want to require tuple or just allow Sequence.

timj:
The other downside of tuple is that:

```python
__all__ = ("myfunc")
```
is not what you want but
```python
__all__ = ["myfunc"]
```
will work without having to worry about the trailing comma.

parejkoj:
Yeah, there's a tradeoff either way. Having `("myfunc")` is caught immediately on import, and we don't have a lot of single-object modules. Should we have a poll?

timj:
I'm happy to let @ktlim decide.

ktlim:
It's really hard to search for `__all__ tuple list` on Slack :)

A GitHub search for `__all__ = (` gives 234K code results; `__all__ = [` gives 963K code results.

The "caught immediately on import" for a missing trailing comma gives a less-than-fully-helpful error (`AttributeError: module 'y' has no attribute 'm'`).

https://stackoverflow.com/questions/66098828/using-list-instead-of-tuple-in-module-all gives some more arguments for list over tuple (including that it is sometimes useful to extend a list programmatically).  "Accidentally" modifying `__all__` seems pretty unlikely.

I would prefer to go with the syntax in the Python docs, so until the docs give a tuple example, we should stick with list.

ktlim:
Note that even cpython is internally inconsistent: the somewhat newer `asyncio` and `tomllib` use tuple, while most other modules seem to use list (including ones where it's required due to modification).

PR #633, opened 2023-08-17
Point to VSCode page in the index

arunkannawadi:


arunkannawadi:
Merging without review after checking that the build is successful and that the added link works as intended.

PR #634, opened 2023-08-25
USDF batch usage hints

jchiang87:


usdf/batch.rst
```diff
@@ -156,6 +167,16 @@ Usage of the ``ctrl_bps_htcondor`` plugin and module has been extensively docume
 
 https://pipelines.lsst.io/modules/lsst.ctrl.bps.htcondor/userguide.html
 
+For running at S3DF, the following ``site`` specification can be used in the BPS configuration file:
```
erykoff, line 170: Thank you for adding this!

usdf/batch.rst
```diff
@@ -4,12 +4,23 @@ Batch Resources
...
+A Slurm job can be submitted with markup ``#SBATCH -p roma,milano`` to run on either of the partitions.
+
+For light interactive work, e.g., running *small* ``pipetask`` jobs, one can obtain an interactive session on the batch nodes using ``srun``.  For example,
+
+.. code-block:: bash
+   :name: srun-interactive-example
+
+   srun --pty --cpus-per-task=4 --mem=6GB --time=02:00:00 --partition=roma,milano --account=rubin bash
+
+will create a 2-hour, 4-core bash session with a total of 6GB memory.  One can then run ``pipetask -j 4``, making use of the 4 allocated cores.  In general, using BPS is preferred to running ``pipetask`` directly since many concurrent ``pipetask`` jobs that are run like this can cause registry database contention.
```
erykoff, line 23: I think we're supposed to add `--nodes=1` to make sure that all our cpus are put on one node (but I don't know what `--pty` does otherwise).  And maybe change the `6GB` to something like `16GB`, the "fair share" for 4 CPUs.  Also, I think it's worth mentioning that specifying the total memory rather than per core is probably what the user wants because otherwise if any process goes over the per-core memory threshold the whole thing gets killed.
And it is also worth mentioning `--exclusive` which will just give you a whole node if that's what you need.

erykoff (approved):
A few small suggestions.

usdf/batch.rst
jchiang87: Thanks, @erykoff .   I've made some changes following your suggestions.  If you give me thumbs up on those, I'll go ahead and merge.

erykoff: Looks fine.  But note that sometimes I just want a whole node to run fgcmcal in debugging mode on a large dataset.  😄

PR #635, opened 2023-09-13
Replace ci.lsst.codes with SLAC URL.

ktlim:
Also update the macOS versions.

### DIFF
--- a/coding/unit-test-policy.rst
+++ b/coding/unit-test-policy.rst
@@ -119,7 +119,7 @@ builds and unit tests expedites the module's repair and, hopefully, limits the
 time other developers are impacted by the failure. For details, refer to the
 workflow documentation on :ref:`workflow-testing`.
 
-.. _Jenkins: https://ci.lsst.codes/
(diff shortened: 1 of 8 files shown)


coding/unit-test-policy.rst
```diff
@@ -119,7 +119,7 @@ builds and unit tests expedites the module's repair and, hopefully, limits the
 time other developers are impacted by the failure. For details, refer to the
 workflow documentation on :ref:`workflow-testing`.
 
-.. _Jenkins: https://ci.lsst.codes/
+.. _Jenkins: https://rubin-ci.slac.stanford.edu/blue
```
mfisherlevine, line 122: One of these has a trailing slash, the other doesn't. I doubt it matters, but it's the only thing that struck me as potentially being wrong.

mfisherlevine (approved):

coding/unit-test-policy.rst
ktlim: It doesn't matter.

PR #637, opened 2023-09-13
Fix broken Jenkins links

arunkannawadi:
@eigerx noted that the Jenkins links on the dev guide are broken. This PR updates the broken instances with the updated link.


arunkannawadi:
Fixed by #635

PR #644, opened 2023-11-01
Update description of test_data path

parejkoj:


usdf/storage.rst
```diff
@@ -22,7 +22,7 @@ Science data under /sdf/group/rubin/:
  - datasets/
  - lsstdata/offline/ (still in prep)
  - repo/
- - shared/data/{test,validation}_data - daily updated checkouts of git lfs repos for use in testing code
+ - shared/data/test_data - daily updated checkouts of git lfs repos for use in testing code (includes the ap_verify CI datasets)
```
ktlim, line 25: You need to add another line for `validation_data`  if you're going to make this just `test_data`.

ktlim (approved):

usdf/storage.rst
parejkoj: We're getting rid of those datasets.

PR #653, opened 2023-12-05
DM-41997 very small docs update

daues:


usdf/batch.rst
```diff
@@ -92,7 +92,8 @@ The ``allocateNodes.py`` utility has the following options::
                         Error log filename; this option for PBS, unused with Slurm
        -g GLIDEINSHUTDOWN, --glidein-shutdown GLIDEINSHUTDOWN
                         glide-in inactivity shutdown time in seconds
-       -v, --verbose         verbose
+       -p, --pack       encouraging nodes to pack jobs rather than spread
```
MichelleGower, line 95: If change code to use "encourage" instead of "encouraging", don't forget to change it here too.

MichelleGower (approved):
Minor comment.  Merge approved.

usdf/batch.rst
daues: I'll change it to 'encourage' to match three other verbs.

PR #662, opened 2024-03-09
Add section on logging in jupyter notebooks

parejkoj:


mfisherlevine:
Changes themselves seem fine. I'd much rather see changes to `~/.ipython/profile_default/startup/20-logging.py` to just have it work though, and same for `python` and `ipython` command line sessions.

mfisherlevine (commented):
Changes themselves seem fine. I'd much rather see changes to `~/.ipython/profile_default/startup/20-logging.py` to just have it work though, and same for `python` and `ipython` command line sessions.

timj:
We can definitely talk about moving the special jupyter logging code into a package outside of lsst-rsp so that it could work in more places.

I think the text should note that if you use the alternative approach then you won't get the same logging output as you would with the lsst-rsp logger override.

parejkoj:
I've tweaked it to note that it's different from the RSP. I agree with Merlin that we should fix the overall logging situation to work in all cases, but that's a bigger question than this ticket.

PR #671, opened 2024-04-23
Fix broken link and cleanup text

parejkoj:


timj (approved):
In some sense sending someone to the pipetask docs when the next sentence sends them to the detailed docs that are applicable to pipetask and butler seems like overkill but I'm fine with it if you are.

PR #675, opened 2024-05-09
DM-43582: Add notes about ci_ap

parejkoj:
Built page: https://developer.lsst.io/v/DM-43582/stack/jenkins-stack-os-matrix.html#running-a-stack-os-matrix-job


timj (approved):
Looks consistent with the other sections.

PR #676, opened 2024-05-15
 encourage astropy.table

womullan:


python/astropy.rst
```diff
@@ -13,7 +13,7 @@ Integration of Astropy core into the LSST software stack is an ongoing process t
 This document is not discussing Astropy affiliated packages, use of which must go through the standard :doc:`RFC </communications/rfc>` process.
 Not all Astropy core packages can be used by default.
 
-The following packages can be used internally in packages if they do not leak into public APIs:
+The following packages can be used internally in packages:
```
ktlim, line 16: I think `time` and `units` are now accepted enough that they could even be in public APIs, not just internal.  But I'm not sure we want the rest in APIs.

python/astropy.rst
```diff
@@ -28,17 +28,20 @@ For reading and writing files in FITS format, both :mod:`astropy.io.fits` and `f
...
 Developers must evaluate which works best for their use case and choose one accordingly.
 
+
 .. _fitsio: https://github.com/esheldon/fitsio
 
 .. warning::
 
     ``lsst.afw.fits`` must not be used in any new Python code, as it is not considered memory safe.
 
+If not interacting with C++ classes the use of ``astropy.table`` is encouraged.
```
ktlim, line 38: Maybe add: "In particular, the ``Table`` class has beneficial properties over Pandas ``DataFrame``." ?

PR #682, opened 2024-06-13
Add info on third-party packages.

ktlim:


stack/packaging-third-party-eups-dependencies.rst
```diff
@@ -2,12 +2,20 @@
...
+This page documents how to make a third-party software package installable using the :command:`eups distrib install` command via two methods: the "forked repo" and the "tarball-and-patch" (TaP) package.
 
-Generally, most dependencies :doc:`should be added via conda </stack/conda>` using the ``rubin-env`` metapackage.
-The primary reason to add an eups tarball-and-patch (TaP) package as described here is if we require a fix that has not been released in conda-forge (or if the package has no conda-forge recipe at all, and we do not want to create one).
+Generally, most dependencies :doc:`should be added via conda </stack/conda>` using the ``rubin-env`` metapackage rather than either of these eups methods.
+If the package is already in conda-forge, it can be added to ``rubin-env`` :ref:`via RFC <third-party-approval>`.
+(Changes to ``rubin-env`` are made approximately 2-4 times a year.)
+If the package is not already in conda-forge, `creating and maintaining a conda-forge recipe`_ is about the same level of effort as maintaining a TaP package, and it makes the package available to the conda-forge community at large rather than just LSST Science Pipelines consumers.
```
taranu, line 10: Maybe it's worth mentioning the rationale for choosing a fork instead here, before the whole fork section? Also because forked repos can be standalone/pip/PyPI installable so the "wider distribution" point is not so much in favour of conda-forge.

stack/packaging-third-party-eups-dependencies.rst
```diff
@@ -22,10 +30,40 @@ See `this page`_ for a list of compatible licenses.
...
+======================
+
+A forked repo relies on building the package from source, rather than using binary artifacts, on the assumption that we may need to use the latest developments.
+Accordingly, we fork the package on GitHub into the ``lsst`` organization and add the "DM Externals" and "Overlords" teams to it as described in :doc:`adding-a-new-package`.
+This ensures that it will be tagged appropriately for official releases (while not conflicting with the third-party tags).
+
+Since the third-party package likely has no support for eups, we need to add a ``ups`` directory.
+This directory should contain the package's EUPS table file as well as an optional file :file:`eupspkg.cfg.sh` which will contain any customized commands for building and installing the third-party package.
+
+The upstream source has no need to know about the ``ups`` directory, so we put it on a separate branch in the ``lsst`` fork.
```
taranu, line 45: I realize that I'm largely alone in this, but if we have some control over upstream, we can add ups support there and just track main, no?

Either way, if a maintainer has access to both, I'm wondering if it's better to put PRs on the fork or upstream first, given the choice.

stack/packaging-third-party-eups-dependencies.rst
```diff
@@ -22,10 +30,40 @@ See `this page`_ for a list of compatible licenses.
...
+That branch is traditionally named ``lsst-dev``.
+We set that branch to be the GitHub default in our fork.
+
+When changes are made to the upstream source, they can be merged into the ``lsst-dev`` branch, or ``lsst-dev`` can be rebased on an upstream branch.
+If we make local changes for our own needs, they can be converted into PRs to the upstream fork, leaving out any commits that refer to the ``ups`` directory contents.
+
+The new package then needs to be added to ``etc/repos.yaml`` in the ``lsst/repos`` repository.
+We indicate that the ``lsst-dev`` branch is the one that we will build from by adding a ``ref:`` clause in that file.
+
+Finally, the package should be added as an eups dependency to some other package or meta-package in its ``.table`` file.
```
taranu, line 55: I'm still slightly confused about whether packages should be added as `setupOptional` or `setupRequired` in `lsst_distrib`. For example, `meas_extensions_multiprofit` will start off being used in an optional pipelineTask but (hopefully) become part of the DRP pipelines. At that point should it be required? I see that other measurement extensions are still optional, but multiprofit is like scarlet in that it's multiband and therefore not actually a measurement plugin. So really I'm wondering if meas_extensions_scarlet ought to be required now even if you can technically still turn it off entirely.

taranu (commented):
Assorted comments.

stack/packaging-third-party-eups-dependencies.rst
ktlim: That's what the very next paragraph was supposed to do.

stack/packaging-third-party-eups-dependencies.rst
ktlim: If you have access to both, it depends on the upstream merge latency (or willingness to have the Science Pipelines `lsst-dev` branch being based on a branch).

stack/packaging-third-party-eups-dependencies.rst
ktlim: For most purposes, there's effectively no difference between `setupOptional` and `setupRequired`.  They only differ in the case where someone has not installed an optional package in their local stack.

If you can do useful things with the depending package (and if its tests, if it has any, pass), then `setupOptional` can be a hint that it's not mandatory to have the dependency.  But the shared stack, CVMFS, stack containers, and the RSP will always have it anyway.

stack/packaging-third-party-eups-dependencies.rst
ktlim: I will invert the phrasing in the topic sentence to make it clear.

parejkoj (commented):
Thanks, one suggestion to put the "you should do this" approach first."

I do still think we should probably remove the TaP approach entirely, since we have no more working examples of it, and I doubt we'll ever need it again.

PR #688, opened 2024-08-02
Update LSST acronym in RST guide

arunkannawadi:
Our own guides should not have the old acronym


jonathansick (approved):

arunkannawadi:
Thanks for the prompt review!

PR #691, opened 2024-08-21
Add information on Kubernetes at USDF.

ktlim:
In particular, add the `module load` commands needed to activate kubectl and vault.

Also add a current list of vclusters.


fritzm (approved):
Looks reasonable to me.  Is it really the case that you just have to re-auth through dex to revive your k8s creds?  I always went all the way through re-executing the copy pasta...

(If you do have to copy/execute, we should probably mention that)

PR #692, opened 2024-08-27
DM-45211: Update shared dataset policies

taranu:


usdf/datasets.rst
```diff
@@ -2,49 +2,65 @@
...
+
+- ``/sdf/group/rubin/datasets`` (henceforth ``/datasets`` for short) is a symlink to ``/sdf/data/rubin/shared/ncsa-datasets``, containing datasets previously stored at NCSA under the now-defunct ``/datasets`` path (i.e. the ``/datasets`` symlink no longer exists at S3DF).
+- ``/sdf/group/rubin/shared`` (henceforth ``/shared`` for short) is a symlink to ``/sdf/data/rubin/shared`` and is the preferred path for new shared datasets, as well as for migrating older datasets.
+- ``/sdf/group/rubin/user`` is a symlink to ``/sdf/data/rubin/user`` and contains user home directories. Shared datasets may reside here temporarily for prototyping but should be moved to ``/shared`` once they start being used by multiple users.
+
+.. _policy_usdf:
+
+Policy
+======
+New shared datasets should be added to ``/shared``, following policies to be detailed in `DM-31704 <https://rubinobs.atlassian.net/browse/DM-31704>`_. Additions or modifications to any dataset that is to be used in regular (re-)processing must have a corresponding RFC. The RFC should contain information about:
```
ktlim, line 24: It's not exactly clear from this whether new shared datasets need an RFC.  We should say what the current status is, even if the policy is up in the air.  If we do require an RFC, then we should make clear that the request for space allocation below is after the RFC has been adopted.

usdf/datasets.rst
```diff
@@ -2,49 +2,65 @@
...
+
+- Description and reason for addition/change/deletion
+- Target top-level-directory for location of addition/change/deletion
+- Organization of data
```
ktlim, line 28: Need the expected size of the dataset (or addition).

usdf/datasets.rst
```diff
@@ -77,67 +93,19 @@ Here is a template for what each refcat's readme should contain:
     The configuration that was used to ingest the data is included in this
     directory as `IngestIndexedReferenceTask.py`.
```
ktlim, line 94: We can actually do something like the immutability/sharing below, as long as the modification access is coming from the interactive or batch nodes (not Kubernetes/notebooks).  But we can put that back in when it's better-defined.

ktlim (approved):

usdf/datasets.rst
taranu: Isn't that basically what's described with setting permissions above?

usdf/datasets.rst
taranu: See the new line. I figure anything getting ingested must have an RFC, and anything being used for regular processing but not in any butler (I'm not sure what that would be, but maybe some camera/summit stuff?) ought to be too. Anything else I suppose doesn't need an RFC, although I'm not sure what the leaves. I guess common git repos like testdata_*?

usdf/datasets.rst
```diff
@@ -2,49 +2,75 @@
...
+- Target top-level-directory for location of addition/change/deletion
+- Organization of data
+- Required disk space
+- Other necessary domain knowledge as identified by project members relating to the contents of the data
+
+External datasets not yet used in regular reprocessing should have a corresponding Jira ticket with similar information.
+
+All newly-added datasets, including external datasets, must follow the guidelines for supplying a :ref:`README <readme_guidelines_usdf>` file. Updates to the readme should be reviewed on subsequent Jira tickets.
+
+Requests for new shared directories for should be sent to ``#ops-usdf-data-curation`` in the Rubin Observatory Slack (lsstc).
```
ktlim, line 41: Let's just make this `usdf-help@slac.stanford.edu` for now.

usdf/datasets.rst
ktlim: Changing ownership is kind of a brute-force means of doing this; we can do something a bit more sophisticated for "temporary unlock to allow modification" or even "permanently allow curation by this group of people other than rubinmgr".

usdf/datasets.rst
```diff
@@ -2,49 +2,76 @@
...
+- Target top-level-directory for location of addition/change/deletion
+- Organization of data
+- Required disk space
+- Other necessary domain knowledge as identified by project members relating to the contents of the data
+
+External datasets not yet used in regular reprocessing should have a corresponding Jira ticket with similar information.
+
+All newly-added datasets, including external datasets, must follow the guidelines for supplying a :ref:`README <readme_guidelines_usdf>` file. Updates to the readme should be reviewed on subsequent Jira tickets.
+
+Requests for new shared directories for should be emailed to ``usdf-help@slac.stanford.edu``.
```
ktlim, line 41:
```suggestion
Requests for new shared directories should be emailed to ``usdf-help@slac.stanford.edu``.
```

ktlim (approved):

PR #693, opened 2024-08-29
DM-23505: Discuss Python package encapsulation

leeskelvin:


ktlim (approved):
This looks good and consistent with the RFC and current best practice.

PR #694, opened 2024-08-29
DM-46049: Require merging external repo changes through a ticket branch

arunkannawadi:


arunkannawadi:
Including an additional commit from Lee based on this conversation: https://lsstc.slack.com/archives/C2B6DQBAL/p1724944376660119

stack/packaging-third-party-eups-dependencies.rst
```diff
@@ -46,7 +46,9 @@ The upstream source has no need to know about the ``ups`` directory, so we put i
 That branch is traditionally named ``lsst-dev``.
 We set that branch to be the GitHub default in our fork.
 
-When changes are made to the upstream source, they can be merged into the ``lsst-dev`` branch, or ``lsst-dev`` can be rebased on an upstream branch.
+The ``main`` or the ``master`` branch of the fork should track the default branch upstream, and can be synced as often as desired.
+When relevant changes are made to the upstream source, they can be merged into the ``lsst-dev`` branch, or ``lsst-dev`` can be rebased on an upstream branch.
```
timj, line 50: I'm trying to understand what we were trying to say here (and it's in the original) since won't rebasing lsst-dev mean that we detach all the w.YYYY.NN tags from the history?

arunkannawadi: OK, my guess here is that rebasing is fine (and may even be preferred) for repos that don't receive weekly tags (metadetect is one such example). Once they are part of `lsst_distrib` or other distributions, they should also get a merge from a ticket branch.

stack/packaging-third-party-eups-dependencies.rst
```diff
@@ -46,7 +46,10 @@ The upstream source has no need to know about the ``ups`` directory, so we put i
 That branch is traditionally named ``lsst-dev``.
 We set that branch to be the GitHub default in our fork.
 
-When changes are made to the upstream source, they can be merged into the ``lsst-dev`` branch, or ``lsst-dev`` can be rebased on an upstream branch.
+The ``main`` or the ``master`` branch of the fork should track the default branch upstream, and can be synced as often as desired.
+When relevant changes are made to the upstream source, they can be merged into the ``lsst-dev`` branch, or ``lsst-dev`` can be rebased on an upstream branch if the repository does not have weekly tags.
+In such cases, rebasing is preferable since it reduces the apparent divergence between ``lsst-dev`` and the default branch upstream.
```
timj, line 51: What we are really saying here is that you can rebase up until the point where you become part of lsst_distrib and the git commit SHA1s become burned into builds. At that point the lsst-dev branch needs to be treated like main from our point of view, including branch protections and disallowing force pushes. If there aren't any changes by us on lsst-dev then it's all fast forward merges from upstream.  Can you remember @ktlim what were we trying to say about rebase here originally?

arunkannawadi: That is exactly my understanding as well. Depending on what @ktlim says, I'd prefer to rebase the changes in metadetect package rather than merging to `lsst-dev`.

ktlim: Devil's advocate: Is there anything particularly wrong with rebasing even after you have weekly tags?  The tags stay, they're just not on a branch.

timj: We might get away with that if we never need to refer to untagged commits. The history relative to HEAD wouldn't show any tags at all but maybe that's fine?

It would definitely make it significantly clearer what changed relative to main. You would still want to make sure merge commits are retained during the rebase.

arunkannawadi: Thanks for the heads up about preserving merge commits. I added a sentence on that as well.

Personally, I'd like to be able to see the tags on GitHub directly, but I think as long as I am able to check out a tagged version (regardless of it being on a branch), I'd be happy. Do I need to add any other clarification to this text?

timj:
I think at this point it would be best for @ktlim to sign off on allowing rebase of forked lsst-dev branches.

arunkannawadi:
Renaming the branch, so I can close off the ticket. Submitting the changes from a user branch now.

PR #695, opened 2024-09-09
Require merging external repo changes through a ticket branch

arunkannawadi:


ktlim (approved):
I'm a little worried that this is kind of fiddly and that `--rebase-merges` might get forgotten, but it's probably the best we can do for now, and the net differences between lsst-dev and the default upstream are always available in the GitHub UI.

PR #699, opened 2024-10-10
Update the Slack usage help channel name

arunkannawadi:
Reflect the new name for the help-slack channel that was changed as per K-T's suggestion.


arunkannawadi:
Hmm, I was pretty sure this needed an explicit approval for auto-merge to merge, but given that the change is trivial, I think it's okay that I YOLO'd it.

PR #700, opened 2024-10-10
Add a note about providing stable interfaces

arunkannawadi:
upon transferring code between packages.


arunkannawadi:
Build webpage is here: https://developer.lsst.io/v/u-arunkannawadi-importability-on-code-transfer/stack/transferring-code.html

PR #723, opened 2025-06-30
DM-51627: Update mentions of cvmfs to almalinux-x86_64 instead of linux-x86_64

airnandez:


usdf/stack.rst
```diff
@@ -8,33 +8,37 @@ LSST Science Pipelines "stack" available at the USDF.
...
 You can see which versions are available by:
-``ls /cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/``
+``ls /cvmfs/sw.lsst.eu/almalinux-x86_64/lsst_distrib/``
 and
-``ls /cvmfs/sw.lsst.eu/containers/apptainer/lsst_distrib/``
+``ls /cvmfs/sw.lsst.eu/containers/apptainer/x86_64/almalinux/lsst_distrib/``
+
+.. note::
+
+	Stable and select weekly releases older than ``w_2025_11``, were built on the CentOS 7 Linux distribution (the previous reference platform). They are available under ``/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib`` (note the ``linux-x86_64`` instead of ``almalinux-x86_64`` in the path).
```
ktlim, line 41:
```suggestion
	Stable and select weekly releases older than ``w_2025_11`` were built on the CentOS 7 Linux distribution (the previous reference platform). They are available under ``/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib`` (note the ``linux-x86_64`` instead of ``almalinux-x86_64`` in the path).
```

usdf/batch.rst
```diff
@@ -321,6 +321,11 @@ to the desired ``site`` subsection.  The ``monitorFilename`` field specifies the
 
   The ``pip uninstall sqlalchemy`` command is needed since the ``pip install 'parsl[monitoring]'`` command installs an earlier version of ``sqlalchemy`` that's incompatible with ``lsst_distrib``.
 
+
```
ktlim, line 324: Should the previous example for Parsl monitoring be updated to a more recent weekly?

Also, I believe that parsl is now available in the shared stack at S3DF/USDF in addition to CVMFS.

ktlim (approved):

usdf/batch.rst
jchiang87: I haven't used that version of the parsl plugin at  USDF in a while, but I can check to see if it still works ok.  if so, i'll update the docs.

airnandez: I slightly reworded so that statement is factually correct without changing the example to a more recent weekly.

PR #728, opened 2025-09-15
SP-2556: add colorblind-friendly plot guidelines to user docs page

plazas:


jonathansick (approved):
Covering graphics in documentation is a great idea. As described in the comment, please update the link to better match the existing document.

PR #732, opened 2025-10-06
Introduce Rust guide

natelust:
This is a first draft of the proposed rust developer guide. All comments and improvement welcome.

### DIFF
--- a/rust/rust.rst
+++ b/rust/rust.rst
@@ -1,0 +1,101 @@
+LSST Rust Development Guide
+===========================
+
+This document outlines the guidelines for developing Rust code within the LSST (Large Synoptic Survey Telescope) DM (Data Management) Stack. It’s designed to complement existing Python and C++ development guides and ensure consistency across the project.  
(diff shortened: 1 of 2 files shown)


rust/rust.rst
```diff
@@ -0,0 +1,109 @@
...
+* As mentioned above, do not arange module structure according to lsst packages. Write modules by related functionality.
+
+11. Build and management system integration
+-------------------------------------------
+
+* Rust packages should be built using maturin, which manages the complexities of compiling and bundling cargo products.  
+* Cargo additionally is to be used manage dependencies and run rust level tests.  
+* pip is used as the mechanism to locally depoly the wheels created by maturin  
+* pytest is used to run python level unit tests  
+* coordinating these scripts for the developer is a Makefile
```
taranu: Is make really the best option here?

natelust: Here yes it is. It is the simplest way to coordinate a few of the backend technologies in a readable way that is also supported by our eupspkg build system already.

